PR #10, opened 2019-11-09
Mark as requiring legacy external storage in manifest

bylaws:
Android 10 and above block access to the legacy java file apis on external storage without this option. In the future the app should move to using the modern SAF apis.


PixelyIon:
Noted, while we move to the newer SAF APIs we'll use this config tweak.

PR #24, opened 2020-01-13
Add a minimal RomFS implementation

rupansh:
romfs.nro works with nce.cpp fixes!


PixelyIon:
We decided to move to creating a dedicated library `libastro` for our Loaders, ROMFS or other FS access in general and therefore this will be no longer required.
We really value your contribution to the emulator but we didn't have much of a clear idea of what exactly we would do at the time, we offer our deepest apologies for the whole situation ❤️

PR #32, opened 2020-02-11
Significantly Improve Several Systems and Emulation Performance

### LINKED ISSUE #27: Move from `ptrace` to Junction Branching
The current system using `ptrace` is fairly slow due to the amount of cycles `ptrace` requires to pass on a syscall to the kernel process, this is primarily due to requiring the kernel as an intermediary for it. Junction branching on the other hand does runs completely in userspace and gives us far more control over everything.

### LINKED ISSUE #28: Introduce Memory Block System and Shared Memory for all allocations
The current memory system depends almost purely on `KPrivateMemory` for proper segmentation. This system didn't work out that well when implementing all memory related SVCs due to having an address as the parameter. So, a system where memory attributes are stored as blocks that can be found with their address would be far better. In addition, if all memory on the guest is shared with the host and works with the previously mentioned memory block system then most memory accesses won't need to use a kernel call rather can directly access the underlying pages.

### LINKED ISSUE #29: Migrate kernel from being single-threaded to a multi-threaded design
The current kernel is heavily handicapped in areas due to it needing to be single threaded, this is suboptimal as it prevents certain design flows like being able to interrupt the control flow of a certain thread due to sleeping while letting the other threads fully function. In addition, there might be a viable boost in performance from this due to the multiple cores in today's processors which would go unused if we stuck to a single-threaded design.

### LINKED ISSUE #30: Add Mutexes and Conditional Variables
The current system improperly implements mutexes, which leads undefined behavior in most cases. Fixing this is required for essentially all multithreaded applications to work correctly. In addition, conditional variables are currently not implemented and are required by a significant amount of multithreaded applications.

### LINKED ISSUE #31: Implement Asynchronous Loading
The current loading mechanism only scans for ROMs while blocking on the UI thread, this is extremely bad design as the UI should remain interactable at all times. This was somewhat acceptable pre-SAF but due to the significant slowdown in searching post-SAF, we need to implement this.

PixelyIon:
This PR makes several changes to the entire emulator as a whole:
* Move from `ptrace` to Junction Branching (#27)
* Introduce Memory Block System and Shared Memory for all allocations (#28)
* Migrate kernel from being single-threaded to a multi-threaded design (#29)
* Add Mutexes and Conditional Variables (#30)
* Implement Asynchronous Loading (#31)

Closes #27, Closes #28, Closes #29, Closes #30, Closes #31


app/src/main/cpp/skyline/kernel/types/KSharedMemory.cpp
```diff
@@ -1,110 +1,163 @@
...
+                    fregs = {
+                        .x0 = block.address,
+                        .x1 = std::min(block.size, (chunk->address + size) - block.address),
+                        .x2 = static_cast<u64>(block.permission.Get()),
+                        .x8 = __NR_mprotect,
+                    };
+                    state.nce->ExecuteFunction(ThreadCall::Syscall, fregs);
+                    if (fregs.x0 < 0)
+                        throw exception("An error occurred while updating private memory's permissions in child process");
                 } else
```
bylaws: Add brackets

app/src/main/cpp/skyline/kernel/types/KSharedMemory.cpp
```diff
@@ -1,110 +1,163 @@
...
+            if (guest.valid() && state.process) {
+                Registers fregs{
+                    .x0 = guest.address,
+                    .x1 = guest.size,
+                    .x8 = __NR_munmap,
+                };
+                state.nce->ExecuteFunction(ThreadCall::Syscall, fregs);
             }
-            procInf.permission = permission;
+        } catch (const std::exception &) {
```
bylaws: ` catch (const std::exception &) {}`?
Do we want to print a warning here?

app/src/main/cpp/skyline/kernel/types/KThread.cpp
```diff
@@ -35,30 +35,4 @@ namespace skyline::kernel::type {
...
-
-    void KThread::WakeUp() {
-        if (status == Status::Sleeping)
-            status = Status::Runnable;
-    }
-
-    void KThread::ClearWaitObjects() {
-        for (auto &object : waitObjects) {
-            auto iter = object->waitThreads.begin();
-            while (iter != object->waitThreads.end()) {
```
bylaws: Would a for loop be better here?

app/src/main/cpp/skyline/kernel/types/KThread.h
```diff
@@ -16,20 +17,15 @@ namespace skyline::kernel::type {
...
-            Sleeping, //!< The thread is sleeping due to svcSleepThread
-            WaitSync, //!< The thread is waiting for a KSyncObject signal
-            WaitMutex, //!< The thread is waiting on a Mutex
-            WaitCondVar, //!< The thread is waiting on a Conditional Variable
-            Runnable, //!< The thread is ready to run after waiting
             Dead //!< The thread is dead and not running
         } status = Status::Created; //!< The state of the thread
-        std::vector<std::shared_ptr<KSyncObject>> waitObjects; //!< A vector holding the objects this thread is waiting for
-        u64 timeout{}; //!< The end of a timeout for svcWaitSynchronization or the end of the sleep period for svcSleepThread
+        std::atomic<bool> cancelSync; //!< This is to flag to a thread to cancel a synchronization call it currently is in
```
bylaws: Think this should be intialised to false, it isn't set in constructor

app/src/main/cpp/skyline/kernel/types/KTransferMemory.cpp
```diff
@@ -1,102 +1,192 @@
...
-        } else if (!process && owner) {
-            state.os->processMap.at(owner)->ReadMemory(reinterpret_cast<void *>(address), cAddress, copySz);
-        } else {
-            throw exception("Transferring from kernel to kernel is not supported");
+        if (mHost && !host) {
+            state.os->memory.DeleteChunk(address);
+            hostChunk = chunk;
+        } else if (!mHost && host)
+            state.os->memory.InsertChunk(chunk);
+        else if (mHost && host)
```
bylaws: Brackets

app/src/main/cpp/skyline/kernel/types/KTransferMemory.cpp
```diff
@@ -1,102 +1,192 @@
...
+            hostChunk = chunk;
+        else if (!mHost && !host) {
```
bylaws: Same as above

app/src/main/cpp/skyline/kernel/types/KTransferMemory.cpp
```diff
@@ -1,102 +1,192 @@
...
+            if (fregs.x0 < 0)
+                throw exception("An error occurred while updating transfer memory's permissions in guest");
+            auto chunk = state.os->memory.GetChunk(address);
+            MemoryManager::InsertBlock(chunk, block);
+        }
     }
 
     KTransferMemory::~KTransferMemory() {
-        if (owner) {
+        if (host)
```
bylaws: Needs brackets

app/src/main/cpp/skyline/nce.cpp
```diff
@@ -1,231 +1,319 @@
...
-        iovec iov = {&registers, sizeof(registers)};
-        long status = ptrace(PTRACE_GETREGSET, pid ? pid : currPid, NT_PRSTATUS, &iov);
-        if (status == -1)
-            throw exception("Cannot read registers, PID: {}, Error: {}", pid, strerror(errno));
+    void NCE::KernelThread(pid_t thread) {
+        try {
+            state.thread = state.process->threads.at(thread);
+            state.ctx = reinterpret_cast<ThreadContext *>(state.thread->ctxMemory->kernel.address);
+            while (true) {
+                asm("yield");
```
bylaws: Do we really want to yield here? Probably an idea to benchmark this first.

app/src/main/cpp/skyline/nce.cpp
```diff
@@ -1,231 +1,319 @@
...
+                    if (Halt)
+                        break;
+                    if (!Surface)
+                        continue;
+                    const u16 svc = static_cast<const u16>(state.ctx->commandId);
+                    try {
+                        if (kernel::svc::SvcTable[svc]) {
+                            state.logger->Debug("SVC called 0x{:X}", svc);
+                            (*kernel::svc::SvcTable[svc])(state);
+                        } else
```
bylaws: Else brackets

app/src/main/cpp/skyline/nce.cpp
```diff
@@ -1,231 +1,319 @@
...
+        } catch (...) {
+            state.logger->Error("An unknown exception has occurred");
+        }
+        if (!Halt) {
+            if (thread == state.process->pid) {
+                jniMtx.lock(GroupMutex::Group::Group2);
+                state.os->KillThread(thread);
+                Halt = true;
+                jniMtx.unlock();
+            } else
```
bylaws: Brackets

app/src/main/cpp/skyline/nce.cpp
```diff
@@ -1,231 +1,319 @@
...
+            regStr += fmt::format("\nStack Pointer: 0x{:X}", ctx->sp);
+        for (u16 index = 0; index < constant::NumRegs - 1; index += 2) {
+            auto xStr = index < 10 ? " X" : "X";
+            regStr += fmt::format("\n{}{}: 0x{:<16X} {}{}: 0x{:X}", xStr, index, ctx->registers.regs[index], xStr, index + 1, ctx->registers.regs[index + 1]);
         }
+        if (numHist) {
+            state.logger->Debug("Process Trace:{}", trace);
+            state.logger->Debug("Raw Instructions: 0x{}", raw);
+            state.logger->Debug("CPU Context:{}", regStr);
+        } else
```
bylaws: Brackets

app/src/main/cpp/skyline/nce.cpp
```diff
@@ -1,231 +1,319 @@
...
+        0xF2E14F80, // MOVK X0, #2684, LSL #48
+        0x9BC07C21, // UMULH X1, X1, X0
+        0xD347FC21, // LSR X1, X1, #7
+        0xD53BE040, // MRS X0, CNTVCT_EL0
+        0x9AC10801, // UDIV X1, X0, X1
+        0x8B010421, // ADD X1, X1, X1, LSL #1
+        0xD37AE420, // LSL X0, X1, #6
+        0xF90003E0, // STR X0, [SP, #0]
+        0xA94107E0, // LDP X0, X1, [SP, #16]
+    };
```
bylaws: Any reason for not including this in guest.S?

app/src/main/cpp/skyline/nce/guest.cpp
```diff
@@ -0,0 +1,318 @@
...
+                "LDR Q2, [SP], #16\n\t"
+                "LDR Q1, [SP], #16\n\t"
+                "LDR Q0, [SP], #16\n\t"
+                "LDP X1, X2, [SP], #16" :: "r"(ctx->registers.x0));
+            return;
+        }
+        while (true) {
+            ctx->state = ThreadState::WaitKernel;
+            while (ctx->state == ThreadState::WaitKernel);
+            if (ctx->state == ThreadState::WaitRun)
```
bylaws: BRackets

app/src/main/cpp/skyline/nce/guest.cpp
```diff
@@ -0,0 +1,318 @@
...
+        }
+    }
+
+    void entry(u64 address) {
+        volatile ThreadContext *ctx;
+        asm("MRS %0, TPIDR_EL0":"=r"(ctx));
+        while (true) {
+            ctx->state = ThreadState::WaitInit;
+            while (ctx->state == ThreadState::WaitInit);
+            if (ctx->state == ThreadState::WaitRun)
```
bylaws: Brackets

app/src/main/cpp/skyline/nce/guest.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+namespace skyline {
+    namespace guest {
```
bylaws: We should add a message to be wary of what functions are used in guest.cpp

app/src/main/cpp/skyline/common.h
```diff
@@ -47,17 +33,17 @@ namespace skyline {
...
         constexpr u32 NroMagic = 0x304F524E; //!< "NRO0" in reverse, this is written at the start of every NRO file
         // NCE
-        constexpr u8 NumRegs = 31; //!< The amount of registers that ARMv8 has
-        constexpr u16 SvcLast = 0x7F; //!< The index of the last SVC
-        constexpr u16 BrkRdy = 0xFF; //!< This is reserved for our kernel's to know when a process/thread is ready
+        constexpr u8 NumRegs = 30; //!< The amount of registers that ARMv8 has
         constexpr u32 TpidrroEl0 = 0x5E83; //!< ID of TPIDRRO_EL0 in MRS
+        constexpr u32 CntfrqEl0 = 0x5F00; //!< ID of CNTFRQ_EL0 in MRS
+        constexpr u32 TegraX1Freq = 19200000; //!< The clock frequency of the Tegra X1 (19.2 MHz)
         constexpr u32 CntpctEl0 = 0x5F01; //!< ID of CNTPCT_EL0 in MRS
```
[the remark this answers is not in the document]
PixelyIon: I considered that but thought I'll do constant moving during the refactor.

app/src/main/cpp/skyline/nce.cpp
PixelyIon: Not really, I suppose it's a good idea to move it in there.

app/src/main/cpp/skyline/nce/guest.h
PixelyIon: Will do.

app/src/main/cpp/skyline/nce.cpp
PixelyIon: We do, this essentially spins doing nothing. That tends to just waste a lot of CPU resources, a yield will redirect energy towards other cores which can utilize it better hopefully.

app/src/main/cpp/skyline/kernel/svc.cpp
```diff
@@ -279,136 +336,150 @@ namespace skyline::kernel::svc {
...
+                state.thread->cancelSync = false;
+                state.ctx->registers.w0 = constant::status::Interrupted;
+                break;
+            }
+            uint index{};
+            for (const auto &object : objectTable) {
+                if (object->signalled) {
+                    state.logger->Debug("svcWaitSynchronization: Signalled handle: 0x{:X}", waitHandles.at(index));
+                    state.ctx->registers.w0 = constant::status::Success;
+                    state.ctx->registers.w1 = index;
```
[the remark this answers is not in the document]
PixelyIon: That wouldn't work ? That'd only increment when the object is signalled ?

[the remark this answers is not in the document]
bylaws: oh oops

app/src/main/cpp/skyline/kernel/types/KThread.cpp
bylaws: :Shrek:

PR #33, opened 2020-02-12
Add npad support for controller input

bylaws:
Introduces an Input class for interfacing with devices in HID shared
memory and an npad controller that uses it.

Needs a ui for binding controllers before merge.


IvarWithoutBones:
Note that this pull request is closed as its being extended and rewritten.

PR #43, opened 2020-04-13
Put a button with a donation link

str3wer:
I want to donate to this fantastic project, but i can't find a donate button, why not put a donation button in the app? ;)


PixelyIon:
Hey there,
I'm assuming you meant to open an issue rather than a pull request. To answer your question, we're currently planning on adding optional ads to the app when we get to releasing it. We might consider donations at some point but we wouldn't want them now as we haven't really delivered the product we want to yet. 

Thanks for your kind words :)

PR #54, opened 2020-06-15
Updates for the Android R SDK and C++ fixes

bylaws:
- Updates the Kotlin code to support the Android R SDK
- Switches the build SDK to Android R
- Enables `-Wall` and fixes the associated warnings
- The warnings for reordering are left disabled as they are rather pedantic


bylaws:
This is will be merged when the R SDK is out of beta.

PR #55, opened 2020-06-20
Rework the Loader system to be shared between C++ and Kotlin

bylaws:
Unifying the loader code will make maintenance easier in the future and ease the process of adding new backends.


app/src/main/cpp/skyline/loader/format.h
```diff
@@ -0,0 +1,13 @@
...
+    /**
+      * @brief This enumerates the types of ROM files
+      * @note This needs to be synchronized with emu.skyline.loader.BaseLoader.RomFormat
+      */
+    enum class RomFormat {
+        NRO, //!< The NRO format: https://switchbrew.org/wiki/NRO
+        XCI, //!< The XCI format: https://switchbrew.org/wiki/XCI
+        NSP, //!< The NSP format from "nspwn" exploit: https://switchbrew.org/wiki/Switch_System_Flaws
+    };
+}
```
PixelyIon, line 13: Can this be relegated to `loader.h`, I think it'd be fairly fitting in there. Thoughts ?

app/src/main/cpp/skyline/loader/nro.h
```diff
@@ -21,40 +24,67 @@ namespace skyline::loader {
...
+
+        /**
+        * @brief This holds various metadata about an NRO, it is only used by homebrew
+        */
+        struct NroAssetHeader {
+            u32 magic; //!< The asset section magic "ASET"
+            u32 version; //!< The format version
+            NroAssetSection icon;
+            NroAssetSection nacp;
+            NroAssetSection romfs;
```
PixelyIon, line 64: These are missing comments on their contents

app/src/main/cpp/skyline/vfs/nacp.h
```diff
@@ -0,0 +1,40 @@
...
+         * @brief This struct encapsulates all the data within an NACP file
+         */
+        struct NacpData {
+            ApplicationTitle titleEntries[0x10]; //!< Title entries for each language
+            u8 _pad_[0x4000 - (0x10 * 0x300)];
+        } nacpContents{};
+        static_assert(sizeof(NacpData) == 0x4000);
+
+      public:
+        NACP(const std::shared_ptr<vfs::Backing> &backing);
```
PixelyIon, line 35: No comment for parameter `backing`

app/src/main/java/emu/skyline/loader/RomFile.kt
```diff
@@ -123,16 +124,81 @@ class AppEntry : Serializable {
...
+     */
+    private external fun destroy(instance : Long)
+
+    /**
+     * This is used to get the [AppEntry] for the specified NRO
+     */
+    fun getAppEntry(uri : Uri) : AppEntry {
+        return if (hasAssets(instance)) {
+            val rawIcon = getIcon(instance)
+            val icon = BitmapFactory.decodeByteArray(rawIcon, 0, rawIcon.size)
```
PixelyIon, line 180: You should only do this if the size isn't 0, otherwise set `icon` to `null`.
```kotlin
val icon = if (rawIcon.size != 0) BitmapFactory.decodeByteArray(rawIcon, 0, rawIcon.size) else null
```

app/src/main/cpp/skyline/loader/nro.cpp
```diff
@@ -1,38 +1,65 @@
...
-    void NroLoader::LoadProcessData(const std::shared_ptr<kernel::type::KProcess> process, const DeviceState &state) {
-        std::vector<u8> text(header.text.size);
-        std::vector<u8> rodata(header.ro.size);
-        std::vector<u8> data(header.data.size);
+    std::vector<u8> NroLoader::GetIcon() {
+        NroAssetSection &segmentHeader = assetHeader.icon;
+        std::vector<u8> buffer(segmentHeader.size);
+
+        backing->Read(buffer.data(), header.size + segmentHeader.offset, segmentHeader.size);
+        return buffer;
```
PixelyIon, line 35: In certain cases the `ASET` section may be missing or the offset/size specified in the `icon` header can be 0. Check [NroLoader.kt](https://github.com/skyline-emu/skyline/blob/4d787c904e12e671c1ff115938b548e79b492130/app/src/main/java/emu/skyline/loader/NroLoader.kt#L21) for reference, you should return a 0-sized vector here in case the icon doesn't exist.

app/src/main/java/emu/skyline/loader/RomFile.kt
bylaws: That's effectively what HasAssets does, all programs with nacp data also have an icon afaik

app/src/main/cpp/skyline/loader/nro.cpp
bylaws: oh right, forgot about that

bylaws: hmm actually this is fine, since assetHeader is zeroed by default and read will accept zero byte reads

PixelyIon: If the `ASET` header isn't there then that part could be filled with random data rather than just 0s.

bylaws: Nope, if asset header is missing then we don't change the assetHeader from the default, which is being filled with 0s.

bylaws: :shrek:

app/src/main/java/emu/skyline/loader/RomFile.kt
PixelyIon: They can have an icon `ASET` header but the icon itself can be empty and the size can be `0`. So, you should add in the check.

PR #56, opened 2020-06-22
Introduce support for RomFS

bylaws:
RomFS is used by homebrew and games to access content files, it is accessed through an IStorage object. This implementation currently only supports the `Read` and `GetSize` IStorage functions which is enough for most homebrew.

### DIFF
--- a/app/src/main/cpp/skyline/common.h
+++ b/app/src/main/cpp/skyline/common.h
@@ -361,6 +361,9 @@ namespace skyline {
     namespace audio {
         class Audio;
     }
+    namespace loader {
+        class Loader;
+    }
 
     /**
      * @brief This struct is used to hold the state of a device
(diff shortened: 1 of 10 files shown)


app/src/main/cpp/skyline/common.h
```diff
@@ -375,6 +378,7 @@ namespace skyline {
         std::shared_ptr<NCE> nce; //!< This holds a reference to the NCE class
         std::shared_ptr<gpu::GPU> gpu; //!< This holds a reference to the GPU class
         std::shared_ptr<audio::Audio> audio; //!< This holds a reference to the Audio class
+        std::shared_ptr<loader::Loader> loader; //!< This holds a reference to the Loader class
```
PixelyIon, line 381: Consider if putting this in `OS` would be a better idea than having it here, if not then keeping it here is fine too.

bylaws: I don't really see much benefit over state, so it's probably best to keep where it is

PR #66, opened 2020-07-04
Implement pl:u for accessing shared fonts

bylaws:
Fonts are stored in an array of TTF data with an 8 byte header
containing a magic and an XOR'd length. Instead of requiring users to
provide original Nintendo fonts we pack open source replacements.
They are generated with the scripts here
https://github.com/FearlessTobi/yuzu_system_archives. All the fonts are
licenced under the Open Font or Apache 2 License so we can include them
all freely.


bylaws:
I originally did have an std::array, however if I do that I can't store each individual in an array or iterate over them due to templates. It is also much easier to generate them automatically as xxd -i can just be used.

bylaws:
Fixed

PR #68, opened 2020-07-07
Extend account services and add support for PopLaunchParameters

bylaws:
PopLaunchParameters is used to obtain the user who started the application.
These are needed for Puyo Puyo Tetris.

### DIFF
--- a/app/src/main/cpp/skyline/services/account/IAccountServiceForApplication.h
+++ b/app/src/main/cpp/skyline/services/account/IAccountServiceForApplication.h
@@ -6,17 +6,30 @@
 #include <services/base_service.h>
 #include <services/serviceman.h>
 
-namespace skyline::service::account {
-    /**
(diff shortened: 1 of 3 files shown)


app/src/main/cpp/skyline/services/account/IAccountServiceForApplication.h
```diff
@@ -6,17 +6,30 @@
...
-     */
-    class IAccountServiceForApplication : public BaseService {
-      public:
-        IAccountServiceForApplication(const DeviceState &state, ServiceManager &manager);
-
+namespace skyline {
+    namespace service::account {
         /**
-         * @brief This provides information about the running application for account services to use (https://switchbrew.org/wiki/Account_services#InitializeApplicationInfoV0)
+         * @brief This hold an accounts user ID
```
PixelyIon, line 12: "This structure holds an account's User ID (https://switchbrew.org/wiki/Account_services#Uid)"

PR #69, opened 2020-07-07
Implement an ADPCM decoder for audren

bylaws:
This is used by many retail games, it only supports mono sound.

This implementation is based on Ryu's.

### DIFF
--- a/app/src/main/cpp/skyline/audio/adpcm_decoder.cpp
+++ b/app/src/main/cpp/skyline/audio/adpcm_decoder.cpp
@@ -1,0 +1,54 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright © 2020 Skyline Team and Contributors (https://github.com/skyline-emu/)
+
+#include <array>
+#include <common.h>
+#include "common.h"
+#include "adpcm_decoder.h"
+
+namespace skyline::audio {
+    AdpcmDecoder::AdpcmDecoder(const std::vector<std::array<i16, 2>> &coefficients) : coefficients(coefficients) {}
+
+    std::vector<i16> AdpcmDecoder::Decode(const std::vector<u8> &adpcmData) {
+        constexpr size_t BytesPerFrame = 0x8; //!< The number of bytes in a single frame
+        constexpr size_t SamplesPerFrame = 0xe; //!< The number of samples in a single frame
+
+        size_t remainingSamples = (adpcmData.size() / BytesPerFrame) * SamplesPerFrame;
+
+        std::vector<i16> output;
+        output.reserve(remainingSamples);
+
+        size_t inputOffset{};
+
+        while (inputOffset < adpcmData.size()) {
+            FrameHeader header(adpcmData[inputOffset++]);
+
+            size_t frameSamples = std::min(SamplesPerFrame, remainingSamples);
+
+            i32 ctx{};
+
+            for (size_t index = 0; index < frameSamples; index++) {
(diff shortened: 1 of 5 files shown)


app/src/main/cpp/skyline/audio/adpcm_decoder.cpp
```diff
@@ -0,0 +1,54 @@
...
+                } else {
+                    ctx = adpcmData[inputOffset++];
+                    sample = (ctx << 24) >> 28;
+                }
+
+                i32 prediction = history[0] * coefficients[header.coefficientIndex][0] + history[1] * coefficients[header.coefficientIndex][1];
+                sample = (sample * (0x800 << header.scale) + prediction + 0x400) >> 11;
+
+                auto saturated = audio::Saturate<i16, i32>(sample);
+                output.emplace_back(saturated);
```
PixelyIon, line 44: Use `push_back` rather than `emplace_back` as it'll fallback to the behavior of the former here regardless

app/src/main/cpp/skyline/audio/adpcm_decoder.h
```diff
@@ -0,0 +1,42 @@
...
+
+            FrameHeader(u8 raw) : raw(raw) {}
+        };
+        static_assert(sizeof(FrameHeader) == 0x1);
+
+        std::array<i32, 2> history{}; //!< This contains the history for decoding the ADPCM stream
+        std::vector<std::array<i16, 2>> coefficients; //!< This contains the coefficients for decoding the ADPCM stream
+
+      public:
+        AdpcmDecoder(const std::vector<std::array<i16, 2>> &coefficients);
```
PixelyIon, line 33: Document the parameter `coefficients`

app/src/main/cpp/skyline/services/audio/IAudioRenderer/voice.cpp
```diff
@@ -60,6 +67,13 @@ namespace skyline::service::audio::IAudioRenderer {
                 samples.resize(currentBuffer.size / sizeof(i16));
                 state.process->ReadMemory(samples.data(), currentBuffer.address, currentBuffer.size);
                 break;
+            case skyline::audio::AudioFormat::ADPCM:
+                {
```
PixelyIon, line 71: The bracket for a case statement should be on the same line at the end of the case statement itself

app/src/main/cpp/skyline/audio/adpcm_decoder.h
bylaws: We don't need to as it's a class member

app/src/main/cpp/skyline/services/audio/IAudioRenderer/voice.cpp
bylaws: Okay, do I end it the same way?

app/src/main/cpp/skyline/audio/adpcm_decoder.h
PixelyIon: I see, my bad

app/src/main/cpp/skyline/services/audio/IAudioRenderer/voice.cpp
PixelyIon: The bracket at the end can be on it's own line, it's just that the starting one has to share the line with the rest of the statement, just like we have it for `if`/`for`/`while` and other bracketed control flow statements.

PR #71, opened 2020-07-09
Further extend our services implementations

bylaws:
This pr implements all the services needed to boot Mario Odyssey, however it will still fail later on due to an unknown bug.

### DIFF
--- a/app/src/main/cpp/skyline/services/am/controller/ISelfController.cpp
+++ b/app/src/main/cpp/skyline/services/am/controller/ISelfController.cpp
@@ -12,6 +12,7 @@ namespace skyline::service::am {
         {0xB, SFUNC(ISelfController::SetOperationModeChangedNotification)},
         {0xC, SFUNC(ISelfController::SetPerformanceModeChangedNotification)},
         {0xD, SFUNC(ISelfController::SetFocusHandlingMode)},
+        {0xE, SFUNC(ISelfController::SetRestartMessageEnabled)},
         {0x10, SFUNC(ISelfController::SetOutOfFocusSuspendingEnabled)},
         {0x28, SFUNC(ISelfController::CreateManagedDisplayLayer)}
     }) {}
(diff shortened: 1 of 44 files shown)


app/src/main/cpp/skyline/services/am/controller/ISelfController.cpp
```diff
@@ -12,6 +12,8 @@ namespace skyline::service::am {
         {0xB, SFUNC(ISelfController::SetOperationModeChangedNotification)},
         {0xC, SFUNC(ISelfController::SetPerformanceModeChangedNotification)},
         {0xD, SFUNC(ISelfController::SetFocusHandlingMode)},
+        {0xE, SFUNC(ISelfController::SetRestartMessageEnabled)},
+
```
PixelyIon, line 16: Extra spacing that isn't required

app/src/main/cpp/skyline/services/pctl/IParentalControlService.cpp
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright © 2020 Skyline Team and Contributors (https://github.com/skyline-emu/)
+
+#include "IParentalControlService.h"
+
+namespace skyline::service::pctl {
+    IParentalControlService::IParentalControlService(const DeviceState &state, ServiceManager &manager) : BaseService(state, manager, Service::pctl_IParentalControlService, "pctl:IParentalControlService", {
+    }) {}
```
PixelyIon, line 8: Sorry, but I just noticed that this wasn't due to line wrapping but was on the next line. We don't tend to put the `}) {}` on the next line if the list is empty, only if it's being terminated while having 1 or more item.

bylaws: Meeeehhhh, it's easier to have it two lines for the future :(

PixelyIon: Fine then, I don't mind it all that much :)

PR #73, opened 2020-07-11
MainActivity: Fix snackbar overlapping FABs

Grarak:


app/src/main/java/emu/skyline/views/CustomLinearLayout.kt
```diff
@@ -0,0 +1,38 @@
...
+package emu.skyline.views
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.View
+import android.widget.LinearLayout
+import androidx.coordinatorlayout.widget.CoordinatorLayout
+import com.google.android.material.snackbar.Snackbar.SnackbarLayout
+
+class CustomLinearLayout : LinearLayout, CoordinatorLayout.AttachedBehavior {
```
PixelyIon, line 15: We'd need a comment here describing the functionality of this class

app/src/main/java/emu/skyline/views/CustomLinearLayout.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    constructor(context : Context) : this(context, null)
+    constructor(context : Context, attrs : AttributeSet?) : this(context, attrs, 0)
+    constructor(context : Context, attrs : AttributeSet?, defStyleAttr : Int) : super(context, attrs, defStyleAttr)
+
+    override fun getBehavior() : CoordinatorLayout.Behavior<CustomLinearLayout> = MoveUpwardBehavior()
+
+    class MoveUpwardBehavior : CoordinatorLayout.Behavior<CustomLinearLayout>() {
```
PixelyIon, line 23: Same as above

app/src/main/java/emu/skyline/views/CustomLinearLayout.kt
```diff
@@ -0,0 +1,38 @@
...
+        override fun layoutDependsOn(parent : CoordinatorLayout, child : CustomLinearLayout, dependency : View) : Boolean {
```
PixelyIon, line 24: This function could be made into a single expression function

app/src/main/java/emu/skyline/views/CustomLinearLayout.kt
```diff
@@ -0,0 +1,38 @@
...
+            return dependency is SnackbarLayout
+        }
+
+        override fun onDependentViewChanged(parent : CoordinatorLayout, child : CustomLinearLayout, dependency : View) : Boolean {
+            val translationY = .0f.coerceAtMost(dependency.translationY - dependency.height)
+            child.translationY = translationY
```
PixelyIon, line 30: You can avoid having the extra variable by directly assigning it to `child.translationY`. I'm assuming you did this for readability but due to the assigned statement already being small enough there isn't a huge point. In addition, `.0f` looks a bit off when combined with a statement, I think doing `(0.0f).coerceAtMost` is far more clear.

Grarak:
I applied the changes you mentioned. Let me know if anything else is not right.

PR #76, opened 2020-07-15
Extend service implementations with newer features

bylaws:
* Uses the device sample rate for audren as it deals with resampling internally, this caused a crash in stardew valley.
* Implements the steady clock which is used in many games.
* Adds some extra stubs that were needed for ARMS
* Extends the language implementation to support post 4.0.0 features
* Fixes an issue where service names that were longer than the maximum length of 8 chars were being read causing them to be missed


app/src/main/cpp/skyline/services/am/controller/ISelfController.h
```diff
@@ -17,6 +17,16 @@ namespace skyline::service::am {
       public:
         ISelfController(const DeviceState &state, ServiceManager &manager);
 
+        /**
+         * @brief This function prevents the running application from being quit via the home button (https://switchbrew.org/wiki/Applet_Manager_services#LockExit)
+         */
+        void LockExit(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response);
+
+        /**
+         * @brief This function allows the running application to be quit via the home button (https://switchbrew.org/wiki/Applet_Manager_services#UnockExit)
```
PixelyIon, line 26: https://switchbrew.org/wiki/Applet_Manager_services#UnlockExit

app/src/main/cpp/skyline/services/timesrv/ISystemClock.h
```diff
@@ -29,5 +29,10 @@ namespace skyline::service::timesrv {
          * @brief This returns the amount of seconds since epoch
          */
         void GetCurrentTime(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response);
+
+        /**
+         * @brief This returns the clock context
```
PixelyIon, line 34: "This returns the **system** clock context"

bylaws:
Addressed feedback

PR #77, opened 2020-07-15
Implement a block based GPU virtual memory manager

bylaws:
The GPU has it's own seperate address space to the CPU. It is able to
address 40 bit addresses and accesses the system memory. A sorted vector
has been used to store blocks as insertions are not very frequent.

### DIFF
--- a/app/src/main/cpp/skyline/gpu/memory_manager.h
+++ b/app/src/main/cpp/skyline/gpu/memory_manager.h
@@ -1,0 +1,101 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright © 2020 Skyline Team and Contributors (https://github.com/skyline-emu/)
+
+#pragma once
+
+#include <common.h>
+
+namespace skyline {
+    namespace constant {
+        constexpr u64 GpuPageSize = 1 << 16; //!< The page size of the GPU address space
+    }
+
+    namespace gpu::vmm {
+        /**
+        * @brief This enumerates the possible states of a memory chunk
+        */
+        enum ChunkState {
+            Unmapped, //!< The chunk is unmapped
+            Allocated, //!< The chunk is allocated but unmapped
+            Mapped //!< The chunk is mapped and a CPU side address is present
+        };
+
+        /**
+        * @brief This describes a chunk of memory and all of it's individual attributes
+        */
+        struct ChunkDescriptor {
+            u64 address; //!< The address of the chunk in the GPU address space
+            u64 size; //!< The size of the chunk in bytes
+            u64 cpuAddress; //!< The address of the chunk in the CPU address space (if mapped)
+            ChunkState state; //!< The state of the chunk
+
+            ChunkDescriptor(u64 address, u64 size, u64 cpuAddress, ChunkState state) : address(address), size(size), cpuAddress(cpuAddress), state(state) {}
+
+            /**
+            * @param chunk The chunk to check
+            * @return If the given chunk can be contained wholly within this chunk
+            */
+            inline bool CanContain(const ChunkDescriptor& chunk) {
+                return (chunk.address >= this->address) && ((this->size + this->address) >= (chunk.size + chunk.address));
+            }
+        };
+
+        /**
+        * @brief The MemoryManager class handles the mapping of the GPU address space
+        */
+        class MemoryManager {
+          private:
+            std::vector<ChunkDescriptor> chunkList; //!< This vector holds all the chunk descriptors
+
+            /**
+             * @brief This finds a chunk of the specified type in the GPU address space that is larger than the given size
+             * @param size The minimum size of the chunk to find
+             * @param state The state desired state of the chunk to find
+             * @return The first unmapped chunk in the GPU address space that fits the requested size
+             */
+            std::optional<ChunkDescriptor> FindChunk(u64 size, ChunkState state);
+
+            /**
+             * @brief This inserts a chunk into the chunk list, resizing and splitting as necessary
+             * @param newChunk The chunk to insert
+             * @return The base virtual GPU address of the inserted chunk
+             */
+            u64 InsertChunk(const ChunkDescriptor &newChunk);
+
+          public:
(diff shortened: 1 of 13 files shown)


app/src/main/cpp/skyline/gpu/memory_manager.h
```diff
@@ -0,0 +1,102 @@
...
+        * @brief This describes a chunk of memory and all of it's individual attributes
+        */
+        struct ChunkDescriptor {
+            u64 address; //!< The address of the chunk in the GPU address space
+            u64 size; //!< The size of the chunk in bytes
+            u64 cpuAddress; //!< The address of the chunk in the CPU address space (if mapped)
+            ChunkState state; //!< The state of the chunk
+
+            /**
+            * @brief Checks if the given chunk can be contained wholly within this chunk
```
PixelyIon, line 33: This `@brief` is redundant and can be elided

app/src/main/cpp/skyline/gpu/memory_manager.h
```diff
@@ -0,0 +1,102 @@
...
+            * @param chunk The chunk to check
+            * @return If the given chunk can be contained wholly within this chunk
+            */
+            inline bool CanContain(const ChunkDescriptor& chunk) {
+                return (chunk.address >= this->address) && ((this->size + this->address) >= (chunk.size + chunk.address));
+            }
+
+            ChunkDescriptor(u64 address, u64 size, u64 cpuAddress, ChunkState state) : address(address), size(size), cpuAddress(cpuAddress), state(state) {}
```
PixelyIon, line 41: We generally make the constructor come right after the public member declarations

app/src/main/cpp/skyline/gpu/memory_manager.h
```diff
@@ -0,0 +1,102 @@
...
+             * @return The first unmapped chunk in the GPU address space that fits the requested size
+             */
+            std::optional<ChunkDescriptor> FindChunk(u64 size, ChunkState state);
+
+            /**
+             * @brief This inserts a chunk into the chunk list, resizing and splitting as necessary
+             * @param newChunk The chunk to insert
+             * @return The base virtual GPU address of the inserted chunk
+             */
+            u64 InsertChunk(const ChunkDescriptor &newChunk);
```
PixelyIon, line 64: The argument could be named just `chunk`, I am aware this is the most pointless nitpick ever

app/src/main/cpp/skyline/services/nvdrv/devices/nvhost_as_gpu.cpp
```diff
@@ -1,8 +1,110 @@
...
+            region.offset = state.gpu->memoryManager.AllocateFixed(region.offset, size);
+        else
+            region.offset = state.gpu->memoryManager.AllocateSpace(size);
+
+        if (region.offset == 0) {
+            state.logger->Warn("Failed to allocate GPU address space region!");
+            buffer.status = NvStatus::BadParameter;
+        }
+
+        state.process->WriteMemory(region, buffer.output[0].address);
```
PixelyIon, line 76: `buffer.output.at(0).address`

app/src/main/cpp/skyline/services/nvdrv/devices/nvhost_as_gpu.cpp
```diff
@@ -1,8 +1,110 @@
...
+            region.offset = state.gpu->memoryManager.MapFixed(region.offset, mapPhysicalAddress, mapSize);
+        else
+            region.offset = state.gpu->memoryManager.MapAllocated(mapPhysicalAddress, mapSize);
+
+        if (region.offset == 0) {
+            state.logger->Warn("Failed to map GPU address space region!");
+            buffer.status = NvStatus::BadParameter;
+        }
+
+        state.process->WriteMemory(region, buffer.output[0].address);
```
PixelyIon, line 108: `buffer.output.at(0).address`

app/src/main/cpp/skyline/gpu/memory_manager.h
bylaws: This isn't the case with other structs

app/src/main/cpp/skyline/gpu/memory_manager.h
bylaws: This breaks my iterator, which is also named chunk... I think I'll keep it as newChunk as it's a bit clearer in the function itself.

app/src/main/cpp/skyline/gpu/memory_manager.h
PixelyIon: That might be because they depend on a function definition for the constructor. In any case, you should do the right thing:tm:.

bylaws: Oh wait no I misread, yeah you're right

app/src/main/cpp/skyline/gpu/memory_manager.h
PixelyIon: You can name things differently in the definition and declaration but that's probably worse than just having it be `newChunk` universally.

bylaws: Yeah, I'll leave it as is

bylaws:
Addressed feedback

app/src/main/cpp/skyline/services/nvdrv/devices/nvhost_as_gpu.cpp
```diff
@@ -1,8 +1,110 @@
...
+            u32 _pad1_;
+
+            struct {
+                u64 offset;
+                u32 page_size;
+                u32 pad;
+                u64 pages;
+            } regions[2];
+        } regionInfo = state.process->GetReference<Data>(buffer.input.at(0).address);
+        state.process->WriteMemory(regionInfo, buffer.output[0].address);
```
PixelyIon, line 46: I missed this one earlier :(

bylaws:
done

PR #96, opened 2020-09-02
Rework Service API to be cleaner with significantly less boilerplate

bylaws:
This patch reduces the burden of adding services significantly, rather than having to create an enum entry and add strings in the constructor it will all be determined at runtime through RTTI. A macro is also used in the service creation case to reduce clutter and further cut down on boilerplate.

### DIFF
--- a/app/src/main/cpp/skyline/services/am/controller/ILibraryAppletCreator.cpp
+++ b/app/src/main/cpp/skyline/services/am/controller/ILibraryAppletCreator.cpp
@@ -11,16 +11,18 @@ namespace skyline::service::am {
         {0xA, SFUNC(ILibraryAppletCreator::CreateStorage)}
     }) {}
 
-    void ILibraryAppletCreator::CreateLibraryApplet(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
+    Result ILibraryAppletCreator::CreateLibraryApplet(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
         manager.RegisterService(SRVREG(ILibraryAppletAccessor), session, response);
+        return {};
     }
 
-    void ILibraryAppletCreator::CreateStorage(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
+    Result ILibraryAppletCreator::CreateStorage(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
         auto size = request.Pop<i64>();
 
         if (size < 0)
             throw exception("Cannot create an IStorage with a negative size");
 
         manager.RegisterService(std::make_shared<IStorage>(state, manager, size), session, response);
+        return {};
     }
 }
--- a/app/src/main/cpp/skyline/services/am/controller/ILibraryAppletCreator.cpp
+++ b/app/src/main/cpp/skyline/services/am/controller/ILibraryAppletCreator.cpp
@@ -6,7 +6,7 @@
 #include "ILibraryAppletCreator.h"
 
 namespace skyline::service::am {
-    ILibraryAppletCreator::ILibraryAppletCreator(const DeviceState &state, ServiceManager &manager) : BaseService(state, manager, Service::am_ILibraryAppletCreator, "am:ILibraryAppletCreator", {
+    ILibraryAppletCreator::ILibraryAppletCreator(const DeviceState &state, ServiceManager &manager) : BaseService(state, manager, {
         {0x0, SFUNC(ILibraryAppletCreator::CreateLibraryApplet)},
         {0xA, SFUNC(ILibraryAppletCreator::CreateStorage)}
     }) {}
--- a/app/src/main/cpp/skyline/services/am/storage/IStorageAccessor.cpp
+++ b/app/src/main/cpp/skyline/services/am/storage/IStorageAccessor.cpp
@@ -12,30 +12,34 @@ namespace skyline::service::am {
         {0xB, SFUNC(IStorageAccessor::Read)}
     }) {}
 
-    void IStorageAccessor::GetSize(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
+    Result IStorageAccessor::GetSize(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
         response.Push<i64>(parent->content.size());
+        return {};
     }
 
-    void IStorageAccessor::Write(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
+    Result IStorageAccessor::Write(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
         auto offset = request.Pop<i64>();
-        auto size = request.inputBuf.at(0).size;
+        auto size = std::min(static_cast<i64>(request.inputBuf.at(0).size), static_cast<i64>(parent->content.size()) - offset);
 
-        if (offset + size > parent->content.size())
-            throw exception("Trying to write past the end of an IStorage");
+        if (offset > parent->content.size())
+            return result::OutOfBounds;
 
-        if (offset < 0)
-            throw exception("Trying to write before the start of an IStorage");
+        if (size > 0)
+            state.process->ReadMemory(parent->content.data() + offset, request.inputBuf.at(0).address, size);
 
-        state.process->ReadMemory(parent->content.data() + offset, request.inputBuf.at(0).address, size);
+        return {};
     }
 
-    void IStorageAccessor::Read(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
-        auto offset = request.Pop<u64>();
-        auto size = request.outputBuf.at(0).size;
+    Result IStorageAccessor::Read(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
+        auto offset = request.Pop<i64>();
+        auto size = std::min(static_cast<i64>(request.inputBuf.at(0).size), static_cast<i64>(parent->content.size()) - offset);
 
-        if (offset + size > parent->content.size())
-            throw exception("Trying to read past the end of an IStorage");
+        if (offset > parent->content.size())
+            return result::OutOfBounds;
 
-        state.process->WriteMemory(parent->content.data() + offset, request.outputBuf.at(0).address, size);
+        if (size > 0)
+            state.process->WriteMemory(parent->content.data() + offset, request.outputBuf.at(0).address, size);
+
+        return {};
     }
 }
--- a/app/src/main/cpp/skyline/services/am/storage/IStorageAccessor.cpp
+++ b/app/src/main/cpp/skyline/services/am/storage/IStorageAccessor.cpp
@@ -6,7 +6,7 @@
 #include "IStorageAccessor.h"
 
 namespace skyline::service::am {
-    IStorageAccessor::IStorageAccessor(const DeviceState &state, ServiceManager &manager, std::shared_ptr<IStorage> parent) : parent(parent), BaseService(state, manager, Service::am_IStorageAccessor, "am:IStorageAccessor", {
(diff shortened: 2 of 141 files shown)


app/src/main/cpp/skyline/services/am/controller/ILibraryAppletCreator.cpp
```diff
@@ -6,21 +6,23 @@
...
         manager.RegisterService(SRVREG(ILibraryAppletAccessor), session, response);
+        return {};
     }
 
-    void ILibraryAppletCreator::CreateStorage(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
+    Result ILibraryAppletCreator::CreateStorage(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
         auto size = request.Pop<i64>();
 
         if (size < 0)
             throw exception("Cannot create an IStorage with a negative size");
```
PixelyIon, line 23: Is there no appropriate error for this?

app/src/main/cpp/skyline/services/am/storage/IStorageAccessor.cpp
```diff
@@ -6,36 +6,40 @@
...
+    Result IStorageAccessor::GetSize(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
         response.Push<i64>(parent->content.size());
+        return {};
     }
 
-    void IStorageAccessor::Write(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
+    Result IStorageAccessor::Write(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
         auto offset = request.Pop<i64>();
-        auto size = request.inputBuf.at(0).size;
+        auto size = std::min(static_cast<i64>(request.inputBuf.at(0).size), static_cast<i64>(parent->content.size()) - offset);
```
PixelyIon, line 22: Why is the size an `i64` rather than `u64`? also replace `size > 0` with `size` in the conditional below with that change.

app/src/main/cpp/skyline/services/am/storage/IStorageAccessor.cpp
```diff
@@ -6,36 +6,40 @@
...
-        state.process->ReadMemory(parent->content.data() + offset, request.inputBuf.at(0).address, size);
+        return {};
     }
 
-    void IStorageAccessor::Read(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
-        auto offset = request.Pop<u64>();
-        auto size = request.outputBuf.at(0).size;
+    Result IStorageAccessor::Read(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
+        auto offset = request.Pop<i64>();
+        auto size = std::min(static_cast<i64>(request.inputBuf.at(0).size), static_cast<i64>(parent->content.size()) - offset);
```
PixelyIon, line 35: Same as above

app/src/main/cpp/skyline/services/audio/IAudioDevice.cpp
```diff
@@ -26,26 +26,32 @@ namespace skyline::service::audio {
...
+    Result IAudioDevice::SetAudioDeviceOutputVolume(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
+        return {};
+    }
 
-    void IAudioDevice::GetActiveAudioDeviceName(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
+    Result IAudioDevice::GetActiveAudioDeviceName(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
         std::string deviceName("AudioStereoJackOutput");
 
         if (deviceName.size() > request.outputBuf.at(0).size)
             throw exception("Too small a buffer supplied to GetActiveAudioDeviceName");
```
PixelyIon, line 40: Is there not an appropriate result for this?

app/src/main/cpp/skyline/services/base_service.h
```diff
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include <cxxabi.h>
```
PixelyIon, line 6: This isn't an STD include per-se, it should come under `<functional>`

app/src/main/cpp/skyline/services/base_service.h
```diff
@@ -138,39 +28,48 @@ namespace skyline::service {
...
-         * @param hasLoop If the service has a loop or not
-         * @param serviceType The type of the service
-         * @param serviceName The name of the service
          * @param vTable The functions of the service
          */
-        BaseService(const DeviceState &state, ServiceManager &manager, Service serviceType, const std::string &serviceName, const std::unordered_map<u32, std::function<void(type::KSession &, ipc::IpcRequest &, ipc::IpcResponse &)>> &vTable) : state(state), manager(manager), serviceType(serviceType), serviceName(serviceName), vTable(vTable) {}
+        BaseService(const DeviceState &state, ServiceManager &manager, const std::unordered_map<u32, std::function<Result(type::KSession &, ipc::IpcRequest &, ipc::IpcResponse &)>> &vTable) : state(state), manager(manager), vTable(vTable) {}
+
+        /**
+         * @brief This is used to enable RTTI for use with service related logging
```
PixelyIon, line 41: `... for use with service related logging` I think `... to be able to extract the name of the underlying class and ensure correct destruction order` is better

app/src/main/cpp/skyline/services/base_service.h
```diff
@@ -138,39 +28,48 @@ namespace skyline::service {
       protected:
         const DeviceState &state; //!< The state of the device
         ServiceManager &manager; //!< A reference to the service manager
-        std::unordered_map<u32, std::function<void(type::KSession &, ipc::IpcRequest &, ipc::IpcResponse &)>> vTable; //!< This holds the mapping from a function's CmdId to the actual function
+        std::unordered_map<u32, std::function<Result(type::KSession &, ipc::IpcRequest &, ipc::IpcResponse &)>> vTable; //!< This holds the mapping from a function's CmdId to the actual function
```
PixelyIon, line 31: We should probably replace `std::function` with a straight function pointer here, we don't use any of the extra functionality provided by it and it's actually slower to my knowledge: `Result(*)(type::KSession &, ipc::IpcRequest &, ipc::IpcResponse &)`.

app/src/main/cpp/skyline/services/fssrv/IFileSystemProxy.cpp
```diff
@@ -61,16 +63,21 @@ namespace skyline::service::fssrv {
...
+        return {};
+
     }
 
-    void IFileSystemProxy::OpenDataStorageByCurrentProcess(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
+    Result IFileSystemProxy::OpenDataStorageByCurrentProcess(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
         if (state.loader->romFs)
             manager.RegisterService(std::make_shared<IStorage>(state.loader->romFs, state, manager), session, response);
         else
             throw exception("Tried to call OpenDataStorageByCurrentProcess without a valid RomFS");
```
PixelyIon, line 74: Is there no appropriate error code for this?

app/src/main/cpp/skyline/services/serviceman.h
```diff
@@ -10,49 +10,46 @@
 namespace skyline::service {
     /**
      * @brief The ServiceManager class manages passing IPC requests to the right Service and running event loops of Services
+     * @todo This implementation varies significantly from HOS, this should be rectified much later on
```
PixelyIon, line 13: Remove this, we don't want such TODOs floating around in the codebase forever

app/src/main/cpp/skyline/services/serviceman.h
```diff
@@ -61,8 +58,13 @@ namespace skyline::service {
...
          */
         template<typename Type>
-        inline std::shared_ptr<Type> GetService(Service serviceType) {
-            return std::static_pointer_cast<Type>(serviceMap.at(serviceType));
+        std::shared_ptr<Type> GetService(ServiceName name) {
+            return std::static_pointer_cast<Type>(serviceMap.at(name));
+        }
+
+        template<typename Type>
+        constexpr std::shared_ptr<Type> GetService(std::string_view name) {
```
PixelyIon, line 66: const ref, a `std::string_view` is still 2 64-bit ints long and shouldn't be copied where possible

app/src/main/cpp/skyline/services/sm/IUserInterface.cpp
```diff
@@ -4,26 +4,28 @@
...
-                manager.NewService(serviceName, session, response);
-            } catch (std::out_of_range &) {
-                response.errorCode = constant::status::ServiceNotReg;
-                state.logger->Warn("Service has not been implemented: \"{}\"", serviceName);
-            }
+        try {
+            manager.NewService(name, session, response);
+        } catch (std::out_of_range &) {
+            std::string_view stringName(reinterpret_cast<char *>(&name), sizeof(u64));
+            state.logger->Warn("Service has not been implemented: \"{}\"", stringName);
```
PixelyIon, line 26: I wonder, should we return `result::InvalidServiceName` here or not?

PixelyIon (changes requested):

app/src/main/cpp/skyline/services/base_service.h
bylaws: We want the name eventually

PixelyIon: I forgot, my bad

app/src/main/cpp/skyline/services/am/storage/IStorageAccessor.cpp
bylaws: Because of the offset potentially being larger than size

app/src/main/cpp/skyline/services/am/controller/ILibraryAppletCreator.cpp
bylaws: This code isn't even part of the pr ;)

app/src/main/cpp/skyline/services/serviceman.h
bylaws: Nah we will need to do it, so it should be kept as it explains our inaccuracies

app/src/main/cpp/skyline/services/sm/IUserInterface.cpp
bylaws: That would be the entirely wrong error code, so no

app/src/main/cpp/skyline/services/serviceman.h
bylaws: The compiler will do this itself, same as std span

app/src/main/cpp/skyline/services/fssrv/IFileSystemProxy.cpp
bylaws: Also not part of the pr :)

app/src/main/cpp/skyline/services/am/controller/ILibraryAppletCreator.cpp
PixelyIon: I'm aware but this is a refactor so mostly anything goes

app/src/main/cpp/skyline/services/am/storage/IStorageAccessor.cpp
bylaws: Actually this kinda needs to be signed, it's the neatest way of doing it.

bylaws: the behaviour would be different

app/src/main/cpp/skyline/services/audio/IAudioDevice.cpp
bylaws: Doesn't seem to be, ryu just returns success

app/src/main/cpp/skyline/services/base_service.h
bylaws: it doesn't

PixelyIon (approved):

PR #97, opened 2020-09-05
Implement Vibration Emulation for Host Vibrators

PixelyIon:
This PR adds in a translation layer for all guest vibration to be converted into corresponding host vibration data and ran on host vibrators using Android Vibration APIs. There's support for using vibrators present in the controllers and those which are present in the device itself.

### DIFF
--- a/app/src/main/cpp/skyline/input/npad_device.cpp
+++ b/app/src/main/cpp/skyline/input/npad_device.cpp
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: MPL-2.0
 // Copyright © 2020 Skyline Team and Contributors (https://github.com/skyline-emu/)
 
+#include <jvm.h>
 #include "npad_device.h"
 #include "npad.h"
 
@@ -154,6 +155,9 @@ namespace skyline::input {
 
         section = {};
         globalTimestamp = 0;
+
+        index = -1;
+        partnerIndex = -1;
 
         type = NpadControllerType::None;
         controllerInfo = nullptr;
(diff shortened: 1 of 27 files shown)


app/src/main/cpp/skyline/input/npad_device.cpp
```diff
@@ -347,4 +351,91 @@ namespace skyline::input {
...
+            jint totalAmplitude{};
+            for (const auto &vibration : vibrations)
+                totalAmplitude += vibration.second;
+
+            if (totalAmplitude == 0 || vibrations[3].first == 0) {
+                manager.state.jvm->ClearVibrationDevice(index);
+                return;
+            }
+
+            for (u8 i{0}; i < vibrations.size(); i++) {
```
bylaws, line 421: i{}

app/src/main/cpp/skyline/input/npad_device.cpp
```diff
@@ -347,4 +351,91 @@ namespace skyline::input {
 
         globalTimestamp++;
     }
+
+    void NpadDevice::VibrateDevice(i8 vibrateIndex, const NpadVibrationValue &value) {
+        std::array<jlong, 3> timings;
```
bylaws, line 356: Make all the magics constants, and explain what each is for

app/src/main/cpp/skyline/input/npad_device.cpp
```diff
@@ -347,4 +351,91 @@ namespace skyline::input {
...
+            vibrationLeft = value;
+
+        if (vibrationRight)
+            Vibrate(vibrationLeft, *vibrationRight);
+        else
+            VibrateDevice(index, value);
+    }
+
+    void NpadDevice::Vibrate(const NpadVibrationValue &left, const NpadVibrationValue &right) {
+        if (partnerIndex == -1) {
```
bylaws, line 397: Same as above

app/src/main/cpp/skyline/services/hid/IHidServer.cpp
```diff
@@ -105,4 +108,34 @@ namespace skyline::service::hid {
...
+        auto &valueBuf = request.inputBuf.at(1);
+        std::span values(reinterpret_cast<NpadVibrationValue *>(valueBuf.address), valueBuf.size / sizeof(NpadVibrationValue));
+
+        for (int i = 0; i < handles.size(); ++i) {
+            auto &handle = handles[i];
+
+            auto &device = state.input->npad.at(handle.id);
+            if (device.type == handle.GetType()) {
+                if (i + 1 != handles.size() && handles[i + 1].id == handle.id && handles[i + 1].isRight && !handle.isRight) {
+                    state.logger->Info("Vibration #{}&{} - Handle: 0x{:02X} (0b{:05b}), Vibration: {:.2f}@{:.2f}Hz, {:.2f}@{:.2f}Hz - {:.2f}@{:.2f}Hz, {:.2f}@{:.2f}Hz", i, i + 1, u8(handle.id), u8(handle.type), values[i].amplitudeLow, values[i].frequencyLow, values[i].amplitudeHigh, values[i].frequencyHigh, values[i + 1].amplitudeLow, values[i + 1].frequencyLow, values[i + 1].amplitudeHigh, values[i + 1].frequencyHigh);
```
bylaws, line 130: Having this as info 100% isn't a good idea, drop it down to debug.

Or alternatively add a new log level so, trace, so there isn't so much SVC garbage in debug

app/src/main/cpp/skyline/services/hid/IHidServer.cpp
```diff
@@ -105,4 +108,34 @@ namespace skyline::service::hid {
...
+
+    void IHidServer::SendVibrationValues(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
+        request.Skip<u64>(); // appletResourceUserId
+
+        auto &handleBuf = request.inputBuf.at(0);
+        std::span handles(reinterpret_cast<NpadDeviceHandle *>(handleBuf.address), handleBuf.size / sizeof(NpadDeviceHandle));
+        auto &valueBuf = request.inputBuf.at(1);
+        std::span values(reinterpret_cast<NpadVibrationValue *>(valueBuf.address), valueBuf.size / sizeof(NpadVibrationValue));
+
+        for (int i = 0; i < handles.size(); ++i) {
```
bylaws, line 124: size_t I{}
Use I++ too

app/src/main/cpp/skyline/services/hid/IHidServer.cpp
```diff
@@ -105,4 +108,34 @@ namespace skyline::service::hid {
...
+            auto &handle = handles[i];
```
bylaws, line 125: const ref for these? I don't see anything writing to them

app/src/main/cpp/skyline/input/npad_device.cpp
```diff
@@ -347,4 +351,91 @@ namespace skyline::input {
...
+
+        jint amplitudeLow = value.amplitudeLow * 127;
+        jint amplitudeHigh = value.amplitudeHigh * 127;
+
+        if (amplitudeLow + amplitudeHigh == 0 || periodLow + periodHigh == 0) {
+            manager.state.jvm->ClearVibrationDevice(vibrateIndex);
+            return;
+        }
+
+        if (periodLow == periodHigh) {
```
bylaws, line 370: Can we have a comment here explaining what's going on?

bylaws (changes requested):

app/src/main/cpp/skyline/services/hid/IHidServer.cpp
PixelyIon: This was meant to be a temporary thing while I debugged, didn't mean to leave it in. I think this is something that'll be solved by our newer logger regardless so it shouldn't be an issue.

app/src/main/cpp/skyline/input/npad_device.h
```diff
@@ -6,6 +6,11 @@
 #include <kernel/types/KEvent.h>
 #include "shared_mem.h"
 
+namespace skyline::constant {
+    constexpr jlong MsInSecond = 1000; //!< The amount of milliseconds in a single second of time
```
bylaws, line 10: I think std::chrono can do this

app/src/main/cpp/skyline/services/hid/IHidServer.cpp
```diff
@@ -105,4 +108,33 @@ namespace skyline::service::hid {
...
+        request.Skip<u64>(); // appletResourceUserId
+
+        auto &handleBuf = request.inputBuf.at(0);
+        std::span handles(reinterpret_cast<NpadDeviceHandle *>(handleBuf.address), handleBuf.size / sizeof(NpadDeviceHandle));
+        auto &valueBuf = request.inputBuf.at(1);
+        std::span values(reinterpret_cast<NpadVibrationValue *>(valueBuf.address), valueBuf.size / sizeof(NpadVibrationValue));
+
+        for (size_t i{}; i < handles.size(); ++i) {
+            const auto &handle = handles[i];
+            auto &device = state.input->npad.at(handle.id);
```
bylaws, line 126: can this be const?

bylaws (changes requested):

app/src/main/cpp/skyline/input/npad_device.h
PixelyIon: It can but I don't think we should explicitly source it from there, similar to how it's done in `common.h` with `NsInSecond` atm

app/src/main/cpp/skyline/services/hid/IHidServer.cpp
PixelyIon: Nope, it can't

app/src/main/cpp/skyline/input/npad_device.h
bylaws: Chrono could probably do it in the code itself, no need for a constant then. Seeing as it can also do casts from an arbitrary tick amount we could use for GPU/device timer too

PixelyIon: I think it actually looks far more unneat with Chrono than it does otherwise, I was using it earlier and explicitly switch to this rather than the other way around. This value is used in the reciprocal and that just doesn't work too well with how `std::chrono` does things and ends up requiring a bunch of casts which looks terrible.

bylaws (approved):

PR #99, opened 2020-09-08
NCA decryption

Grarak:
This adds support for NCA decryption with mbedtls.

- Import title keys and prod keys in settings
- Decryption of NCA header, Pfs0 and RomFS
- No BKTR support
- No ticket derivation

Old PR: https://github.com/skyline-emu/skyline/pull/82 (was closed by accident)


app/src/main/cpp/skyline/crypto/aes_cipher.cpp
```diff
@@ -0,0 +1,71 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright © 2020 Skyline Team and Contributors (https://github.com/skyline-emu/)
+
+#include "aes_cipher.h"
+
+namespace skyline::crypto {
+    AesCipher::AesCipher(const std::span<u8> &key, mbedtls_cipher_type_t type) {
```
bylaws, line 7: No need to pass spans by reference, the compiler will optimise passing them

app/src/main/cpp/skyline/crypto/aes_cipher.h
```diff
@@ -0,0 +1,73 @@
...
+
+        /**
+         * @note destination and source can be the same
+         */
+        void Decrypt(u8 *destination, u8 *source, size_t size);
+
+        /**
+         * @brief Decrypts data and writes back to it
+         */
+        inline void Decrypt(const std::span<u8> &data) {
```
bylaws, line 57: No need to pass spans by reference

app/src/main/cpp/skyline/crypto/aes_cipher.h
```diff
@@ -0,0 +1,73 @@
...
+
+        /**
+         * @brief Decrypts data with XTS. IV will get calculated with the given sector
+         */
+        void XtsDecrypt(u8 *destination, u8 *source, size_t size, size_t sector, size_t sectorSize);
+
+        /**
+         * @brief Decrypts data with XTS and writes back to it
+         */
+        inline void XtsDecrypt(const std::span<u8> &data, size_t sector, size_t sectorSize) {
```
bylaws, line 69: No need to pass spans by reference,  also make this templated and use span<T> , same for others

app/src/main/cpp/skyline/crypto/key_store.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+            std::string_view key(&*lineStart, keyEnd - lineStart);
+            std::string_view value(&*(keyEnd + 1), lineEnd - keyEnd - 1);
+            (this->*callback)(key, value);
+
+            lineStart = lineEnd + 1;
+        }
+    }
+
+    void KeyStore::PopulateTitleKeys(const std::string_view &keyName, const std::string_view &value) {
```
bylaws, line 37: String views don't need to be passed by reference either I'm fairly sure, though check godbolt with O3 on that

Grarak: Yeah, there is no difference on godbolt

app/src/main/cpp/skyline/crypto/key_store.cpp
```diff
@@ -0,0 +1,75 @@
...
+                        return;
+                    case Keys128::KeyAreaKeyOcean:
+                        areaKeyOcean[index] = key;
+                        return;
+                    case Keys128::KeyAreaKeySystem:
+                        areaKeySystem[index] = key;
+                        return;
+                }
+            }
+        }
```
PixelyIon, line 73: Eliminate `Keys128` by using `offsetof` to the members instead, it's redundant to have it be converted into an unrelated enum in the middle. You can also make the `unordered_map` into a `map` to search for any prefixes directly: https://stackoverflow.com/a/57905386.

app/src/main/cpp/skyline/crypto/key_store.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+    void KeyStore::PopulateKeys(const std::string_view &keyName, const std::string_view &value) {
+        auto it{key256Names.find(keyName)};
+        if (it != key256Names.end()) {
+            switch (it->second) {
+                case Keys256::Header:
+                    headerKey = util::HexStringToArray<32>(value);
+                    return;
+            }
+        }
```
PixelyIon, line 51: Same, eliminate `Keys256` by encoding the offset of members instead

app/src/main/cpp/skyline/vfs/ctr_encrypted_backing.cpp
```diff
@@ -0,0 +1,48 @@
...
+// Copyright © 2020 Skyline Team and Contributors (https://github.com/skyline-emu/)
+
+#include "ctr_encrypted_backing.h"
+
+namespace skyline::vfs {
+    constexpr size_t SectorSize = 0x10;
+
+    CtrEncryptedBacking::CtrEncryptedBacking(crypto::KeyStore::Key128 &ctr, crypto::KeyStore::Key128 &key, const std::shared_ptr<Backing> &backing, size_t baseOffset) : Backing({true, false, false}), ctr(ctr), cipher(key, MBEDTLS_CIPHER_AES_128_CTR), backing(backing), baseOffset(baseOffset) {}
+
+    void CtrEncryptedBacking::UpdateCtr(size_t offset) {
```
PixelyIon, line 11: You might want to explicitly use `u64` here, not `size_t`

app/src/main/cpp/skyline/vfs/partition_filesystem.cpp
```diff
@@ -19,12 +19,12 @@ namespace skyline::vfs {
         size_t stringTableOffset = sizeof(FsHeader) + (header.numFiles * entrySize);
         fileDataOffset = stringTableOffset + header.stringTableSize;
 
-        std::vector<char> stringTable(header.stringTableSize);
+        std::vector<char> stringTable(header.stringTableSize + 1);
         backing->Read(stringTable.data(), stringTableOffset, header.stringTableSize);
```
PixelyIon, line 23: You should explicitly do `stringTable[header.stringTableSize] = 0;`. `std::vector` doesn't 0-initialize by default, nor do we want to do that here as most of it is overwritten.

PixelyIon (changes requested):
A few final changes :)

app/src/main/cpp/skyline/vfs/partition_filesystem.cpp
Grarak: `stringTable[header.stringTableSize] = 0;` doesn't work. It throws an error `Variable-sized object may not be initialized`. Should I just memset the vector with `0` or is there a better alternative?

PixelyIon (approved):
LGTM! Great job :)

PR #104, opened 2020-09-27
On-Screen Controls

Grarak:
Adds basic On-Screen Controls when Handheld + Pro Controller is selected as the first controller.
- Resizing
- Layout editing
- L3/R3 stick pressing when double tapping
- Visibility toggling of each button


PixelyIon (approved):
I really have nothing to say, you're just better than me at Kotlin. The single review comment that I have is just grammatical, you can fix that in the UI rework PR rather than this one to avoid rebasing.

PR #107, opened 2020-10-03
Extend NvServices and implement IDirectory

bylaws:
IDirectory is used by many retail games to list the contents of directories, the NvServices functions added are all used by > 3.0.0 games.

### DIFF
--- a/app/src/main/cpp/skyline/services/nvdrv/devices/nvhost_as_gpu.cpp
+++ b/app/src/main/cpp/skyline/services/nvdrv/devices/nvhost_as_gpu.cpp
@@ -30,7 +30,7 @@ namespace skyline::service::nvdrv::device {
         if (region.flags & 1)
             region.offset = state.gpu->memoryManager.ReserveFixed(region.offset, size);
         else
-            region.offset = state.gpu->memoryManager.ReserveSpace(size);
+            region.offset = state.gpu->memoryManager.ReserveSpace(size, region.align);
 
         if (region.offset == 0) {
             state.logger->Warn("Failed to allocate GPU address space region!");
--- a/app/src/main/cpp/skyline/services/nvdrv/devices/nvhost_as_gpu.cpp
+++ b/app/src/main/cpp/skyline/services/nvdrv/devices/nvhost_as_gpu.cpp
@@ -7,6 +7,14 @@
 #include "nvhost_as_gpu.h"
 
 namespace skyline::service::nvdrv::device {
+    struct MappingFlags {
+        bool fixed : 1;
+        u8 _pad0_ : 7;
+        bool remap : 1;
+        u32 _pad1_ : 23;
+    };
+    static_assert(sizeof(MappingFlags) == sizeof(u32));
+
     NvHostAsGpu::NvHostAsGpu(const DeviceState &state) : NvDevice(state) {}
 
     NvStatus NvHostAsGpu::BindChannel(IoctlType type, span<u8> buffer, span<u8> inlineBuffer) {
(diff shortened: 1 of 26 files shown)


app/src/main/cpp/skyline/vfs/directory.h
```diff
@@ -19,17 +22,19 @@ namespace skyline::vfs {
         struct Entry {
             std::string name;
             EntryType type;
+            size_t size; //!< 0 if a directory
         };
 
         /**
          * @brief A descriptor for what will be returned when reading a directories contents
+         * @note MUST match HOS
```
PixelyIon, line 30: This doesn't fit the rest of the codebase, a lot of structures need to match HOS and we don't explicitly mark them as such

app/src/main/cpp/skyline/services/nvdrv/devices/nvhost_as_gpu.cpp
```diff
@@ -43,32 +51,68 @@ namespace skyline::service::nvdrv::device {
...
 
+            if (data.flags.remap) {
+                auto region{regionMap.upper_bound(data.offset)};
+                if ((region == regionMap.begin()) || (region == regionMap.end())) {
+                    state.logger->Warn("Cannot remap an unmapped GPU address space region: 0x{:X}", data.offset);
+                    return NvStatus::BadParameter;
+                }
+
+                // Upper bound gives us the region after the one we want
+                region--;
```
PixelyIon, line 94: Move the comment to the end of the line

app/src/main/cpp/skyline/services/nvdrv/devices/nvhost_ctrl_gpu.cpp
```diff
@@ -77,7 +77,13 @@ namespace skyline::service::nvdrv::device {
         if (data.gpuCharacteristicsBufSize < sizeof(GpuCharacteristics))
             return NvStatus::InvalidSize;
 
-        data.gpuCharacteristics = GpuCharacteristics{};
+        // The ioctl3 version of GetCharacteristics additionally outputs to the inline output buffer
```
PixelyIon, line 80: `ioctl3` -> `IOCTL3`

app/src/main/cpp/skyline/vfs/directory.h
```diff
@@ -19,17 +22,19 @@ namespace skyline::vfs {
...
          */
         union ListMode {
             struct {
                 bool directory : 1; //!< The directory listing will contain subdirectories
                 bool file : 1; //!< The directory listing will contain files
             };
-            u32 raw;
+            u32 raw{};
```
PixelyIon, line 37: This doesn't affect anything unless you move it to above the `struct`

PixelyIon (changes requested):

app/src/main/cpp/skyline/vfs/directory.h
bylaws: Well this is part of vfs, other vfs structs don't need to match hos (see entry below). If you have a list of structs in other subsystems that need to match hos tell me and I'll add notes

app/src/main/cpp/skyline/vfs/directory.h
bylaws: nope, it does, it's just that unions only zero the first entry by default with {}. Adding {} makes it also zero out that as we discussed ages back. I did it this way as it allows you to use ListMode{} neatly.

app/src/main/cpp/skyline/vfs/directory.h
PixelyIon: This applies to a lot of subsystems, it's readily apparent when structures (Esp. unions) are meant for the guest. This shouldn't just set aside that entire convention and need one here.

bylaws: Not really, anyone could assume it's not meant for the guest. Please tell me what subsystems you speak of so I can fix convention

app/src/main/cpp/skyline/vfs/directory.h
PixelyIon: You're wrong, the entire issue was that regardless of what you do it'll only init the first member in unions. You will need to change the order if you want the entire u32 to be 0-init'd.

bylaws: You sure, I don't remember discussing that at all?

app/src/main/cpp/skyline/vfs/directory.h
PixelyIon: The convention doesn't need fixing but essentially all services, kernel objects and GPU.

bylaws: all of them are obvious and all structs there are almost always guest, that isn't the case for vfs where most aren't for HOS

PixelyIon: Are you saying this isn't obvious with it being a union with a `raw` value and a bitfield struct, if you assume this is not for guest then I'm not sure what to say

bylaws: it's definitely not obvious for EntryType, it's literally an enum

PixelyIon (approved):

PR #109, opened 2020-11-03
NCE3, VMM Rework, Exceptional Signal Handler & More

PixelyIon:
The following items were implemented in this PR:
* NCE3 (SPGE) - Same Process Guest Execution
* NCE Pre-patching
* NCE Main Thread Guest Execution
* CPU VMM Rework
* Graceful Exiting using Exceptional Signal Handler
* SVC Implementations:
* * `GetThreadCoreMask`
* * `SetThreadCoreMask`
* * `GetCurrentProcessorNumber`
* * `MapPhysicalMemory`
* * `UnmapPhysicalMemory`
* SVC Improvements:
* * `CreateThread` (Use Core ID)
* * `ExitProcess` & `ExitThread` (Due to Exceptional Signal Handler)
* * `GetInfo` (NPDM Data + Accurate Memory Info)
* * `QueryMemory` (Due to new VMM)
* * `UnmapMemory` (Fix Unmap Bug)
* Rework `KProcess` + `KThread` + `KMemory` derivatives
* Implement `CircularQueue` for efficient GPFIFO operation
* `Verbose` Log Level + Use Android Log API + Thread Tags


app/src/main/cpp/skyline/loader/nca.cpp
```diff
@@ -13,19 +11,22 @@ namespace skyline::loader {
...
-    void NcaLoader::LoadExeFs(const std::shared_ptr<vfs::FileSystem> &exeFs, const std::shared_ptr<kernel::type::KProcess> process, const DeviceState &state) {
+    void* NcaLoader::LoadExeFs(const std::shared_ptr<vfs::FileSystem> &exeFs, const std::shared_ptr<kernel::type::KProcess> process, const DeviceState &state) {
         if (exeFs == nullptr)
             throw exception("Cannot load a null ExeFS");
 
         auto nsoFile{exeFs->OpenFile("rtld")};
         if (nsoFile == nullptr)
             throw exception("Cannot load an ExeFS that doesn't contain rtld");
 
+        state.process->memory.InitializeVmm(memory::AddressSpaceType::AddressSpace39Bit);
```
bylaws, line 22: Uhhh.
Use npdm :)

app/src/main/cpp/skyline/nce.h
```diff
@@ -6,77 +6,32 @@
...
         /**
-         * @brief Patches specific parts of the code
-         * @param code A vector with the code to be patched
-         * @param baseAddress The address at which the code is mapped
-         * @param offset The offset of the code block from the base address
+         * @brief Writes the .patch section and mutates the code accordingly
+         * @param patch A pointer to the .patch section which should be exactly patchSize in size and located before the .text section
          */
-        std::vector<u32> PatchCode(std::vector<u8> &code, u64 baseAddress, i64 offset);
+        static void PatchCode(std::vector<u8> &text, u32* patch, size_t patchSize, const std::vector<size_t>& offsets);
```
bylaws, line 35: Can you run the formatter, we have a bunch of pointers and references that are on the wrong side?

app/src/main/cpp/skyline/nce/guest.h
```diff
@@ -3,45 +3,127 @@
...
-        constexpr size_t LoadCtxSize{20 * sizeof(u32)}; //!< The size of the LoadCtx function in 32-bit ARMv8 instructions
-        constexpr size_t RescaleClockSize{16 * sizeof(u32)}; //!< The size of the RescaleClock function in 32-bit ARMv8 instructions
-        #ifdef NDEBUG
-        constexpr size_t SvcHandlerSize{225 * sizeof(u32)}; //!< The size of the SvcHandler (Release) function in 32-bit ARMv8 instructions
-        #else
-        constexpr size_t SvcHandlerSize{400 * sizeof(u32)}; //!< The size of the SvcHandler (Debug) function in 32-bit ARMv8 instructions
-        #endif
+    struct DeviceState;
+    namespace constant {
+        constexpr u64 SkyTlsMagic{util::MakeMagic<u64>("SKYTLS")};
```
bylaws, line 11: Surely this can be more magical, we have two chars left!

app/src/main/cpp/skyline/services/am/controller/IWindowController.cpp
```diff
@@ -8,7 +8,7 @@ namespace skyline::service::am {
     IWindowController::IWindowController(const DeviceState &state, ServiceManager &manager) : BaseService(state, manager) {}
 
     Result IWindowController::GetAppletResourceUserId(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
-        response.Push(static_cast<u64>(state.process->pid));
+        response.Push(static_cast<u64>(0));
```
bylaws, line 11: Why this change?

bylaws (changes requested):

app/src/main/cpp/emu_jni.cpp
```diff
@@ -7,35 +7,25 @@
...
-    std::signal(SIGINT, signalHandler);
-    std::signal(SIGILL, signalHandler);
-    std::signal(SIGABRT, signalHandler);
-    std::signal(SIGFPE, signalHandler);
+    Fps = 0;
+    FrameTime = 0;
 
-    setpriority(PRIO_PROCESS, static_cast<id_t>(gettid()), -8); // Set the priority of this process to the highest value
+    pthread_setname_np(pthread_self(), "EmuMain");
+    setpriority(PRIO_PGRP, static_cast<id_t>(gettid()), -8); // Set the priority of this process to the highest value
```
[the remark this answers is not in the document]
PixelyIon: This is for the entire Skyline emulator process group, I've removed the guest -> host priority thing as I think it's pointless because the same semantics don't apply.

app/src/main/cpp/skyline/kernel/memory.cpp
```diff
@@ -5,151 +5,105 @@
...
-
-                        iter->size = block.address - iter->address;
-                        chunk->blockList.insert(std::next(iter), {block, endBlock});
-                    }
-                    return;
-                }
+        // Search for a suitable carveout in host AS to fit the guest AS inside of
+        std::ifstream mapsFile("/proc/self/maps");
+        std::string maps((std::istreambuf_iterator<char>(mapsFile)), std::istreambuf_iterator<char>());
+        size_t line{}, start{1ULL << 35}, alignedStart{1ULL << 35}; // 1 << 35 is where QC KGSL (Kernel Graphic Support Layer) maps down from, we skip over this or KGSL goes OOM
```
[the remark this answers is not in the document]
PixelyIon: I don't at all find the specificity erroneous

app/src/main/cpp/emu_jni.cpp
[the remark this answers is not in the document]
bylaws: don't we still have rescaling stuff?

But yeah I would keep it at default then, setting high priority unnecessarily starves other processes

app/src/main/cpp/skyline/kernel/memory.cpp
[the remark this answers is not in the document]
bylaws: I do, it is pointless, we aren't a GPU drivee

app/src/main/cpp/skyline/kernel/memory.cpp
```diff
@@ -5,151 +5,105 @@
...
-    BlockDescriptor *MemoryManager::GetBlock(u64 address, ChunkDescriptor *chunk) {
-        if (!chunk)
-            chunk = GetChunk(address);
+    constexpr size_t RegionAlignment{1ULL << 21}; //!< The minimum alignment of a HOS memory region
 
-        if (chunk) {
-            auto block{std::upper_bound(chunk->blockList.begin(), chunk->blockList.end(), address, [](const u64 address, const BlockDescriptor &block) -> bool {
-                return address < block.address;
-            })};
+    void MemoryManager::InitializeVmm(memory::AddressSpaceType type) {
```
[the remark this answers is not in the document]
PixelyIon: That's also a function, they're different

app/src/main/cpp/skyline/loader/nca.cpp
PixelyIon: Ah, I forgot to do this after implementing NPDM

app/src/main/cpp/skyline/services/am/controller/IWindowController.cpp
PixelyIon: We don't have a PID any longer, only TIDs of individual threads.

app/src/main/cpp/emu_jni.cpp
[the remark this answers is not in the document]
PixelyIon: This is recommended behaviour for the foreground application (Android will automatically reset this when it's not in focus iirc), I would also set the core mask to the high performance cores but we'd want our worker threads to potentially be scheduled on small cores.

app/src/main/cpp/skyline/common.cpp
```diff
@@ -133,36 +21,52 @@ namespace skyline {
...
 
         std::lock_guard guard(mtx);
         logFile << "0|" << str << "\n";
+        logFile.flush();
     }
 
     void Logger::Write(LogLevel level, std::string str) {
-        constexpr std::array<char, 4> levelCharacter{'0', '1', '2', '3'}; // The LogLevel as written out to a file
-        constexpr std::array<int, 4> levelAlog{ANDROID_LOG_ERROR, ANDROID_LOG_WARN, ANDROID_LOG_INFO, ANDROID_LOG_DEBUG}; // This corresponds to LogLevel and provides it's equivalent for NDK Logging
+        constexpr std::array<char, 5> levelCharacter{'0', '1', '2', '3', '4'}; // The LogLevel as written out to a file
```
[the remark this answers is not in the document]
PixelyIon: Not local constants, the regular variable naming convention applies to them.

app/src/main/cpp/skyline/common.h
```diff
@@ -104,14 +131,27 @@ namespace skyline {
...
+        size_t PointerValue(T *item) {
+            return reinterpret_cast<uintptr_t>(item);
+        }
+
         /**
          * @return The value aligned up to the next multiple
          * @note The multiple needs to be a power of 2
          */
         template<typename TypeVal, typename TypeMul>
         constexpr TypeVal AlignUp(TypeVal value, TypeMul multiple) {
```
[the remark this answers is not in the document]
PixelyIon: I disagree, you should see it's use cases as it's used to convert a pointer to a value to temporarily get integral arithmetic properties. This is used in cases like aligning pointers, you'd not want to impose a limitation of purely pointers to integrals.

app/src/main/cpp/skyline/common.h
```diff
@@ -165,28 +207,46 @@ namespace skyline {
...
-                result[i] = (HexDigitToByte(hexString[hexStrIndex]) << 4) | HexDigitToByte(hexString[hexStrIndex + 1]);
+                size_t index{i * 2};
+                result[i] = (HexDigitToNibble(string[index]) << 4) | HexDigitToNibble(string[index + 1]);
             }
             return result;
         }
 
+        template<class Type>
+        constexpr Type HexStringToInt(std::string_view string) {
+            Type result{};
```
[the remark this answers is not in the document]
PixelyIon: `is_integral` is entirely unnecessary, the error generated for if it's a non-integral is fairly readable:
```cpp
HexStringToInt<std::string>("123456789ABCDEF")
```
```
<source>:21:23: error: no matching conversion for static_cast from 'int' to 'std::basic_string<char>'
            result |= static_cast<Type>(digit - '0') << offset;
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

app/src/main/cpp/skyline/common/circular_queue.h
```diff
@@ -0,0 +1,117 @@
...
+        std::vector<u8> vector; //!< The internal vector holding the circular queue's data, we use a byte vector due to the default item construction/destruction semantics not being appropriate for a circular buffer
+        Type *start{reinterpret_cast<Type *>(vector.begin().base())}; //!< The start/oldest element of the queue
+        Type *end{reinterpret_cast<Type *>(vector.begin().base())}; //!< The end/newest element of the queue
+        std::mutex consumptionMutex;
+        std::condition_variable consumeCondition;
+        std::mutex productionMutex;
+        std::condition_variable produceCondition;
+
+      public:
+        CircularQueue(size_t size) : vector(size * sizeof(Type)) {}
```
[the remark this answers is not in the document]
PixelyIon: We don't need to, our operators correctly construct and destroy the underlying types. I've implemented a proper move constructor and explicitly disabled the copy constructor, it was implicitly disabled already due to `std::mutex` and `std::conditional_variable` but being more explicit is better.

app/src/main/cpp/skyline/kernel/memory.cpp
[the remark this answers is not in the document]
PixelyIon: *Looks at nvdrv*

app/src/main/cpp/skyline/kernel/memory.cpp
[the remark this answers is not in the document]
PixelyIon: The issue is the ambiguity created when using generic terms like `AddressSpace` as we have two functions which could be called that, `InitializeVmm` and `InitializeRegions` where one allocates the VMM carveout in the host AS and the latter initializes the regions to define the guest AS.

app/src/main/cpp/skyline/kernel/memory.cpp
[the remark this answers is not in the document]
bylaws: GPU userspace API driver*

app/src/main/cpp/skyline/kernel/memory.cpp
```diff
@@ -160,65 +114,99 @@ namespace skyline::kernel {
...
-                }
-            };
         }
+    }
+
+    std::optional<ChunkDescriptor> MemoryManager::Get(void *ptr) {
+        std::shared_lock lock(mutex);
+
+        auto chunk{std::upper_bound(chunks.begin(), chunks.end(), reinterpret_cast<u8 *>(ptr), [](const u8 *ptr, const ChunkDescriptor &chunk) -> bool { return ptr < chunk.ptr; })};
+        if (chunk-- != chunks.begin())
```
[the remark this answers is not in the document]
PixelyIon: I think we've already discussed this but it's a post-decrement, there's no case where things break as the value is only used if `chunk` isn't at the beginning.

app/src/main/cpp/skyline/kernel/svc.cpp
```diff
@@ -21,241 +21,261 @@ namespace skyline::kernel::svc {
...
-        state.os->KillThread(state.thread->tid);
+    void ExitThread(const DeviceState &state) {
+        state.logger->Debug("svcExitThread: Exiting current thread: {}", state.thread->id);
+        std::longjmp(state.thread->originalCtx, true);
     }
 
-    void SleepThread(DeviceState &state) {
-        auto in{state.ctx->registers.x0};
+    void SleepThread(const DeviceState &state) {
+        u64 in{state.ctx->gpr.x0};
```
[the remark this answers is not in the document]
PixelyIon: I've changed all instances of a register value being taken as a `u64` into explicitly `u64` rather than `auto` as I think `auto` is just longer, it also aligns with using `u32` for all `WRegister` values because it calls `WRegister::operator u32()` for it.

app/src/main/cpp/skyline/common/settings.h
```diff
@@ -0,0 +1,50 @@
...
+         * @param fd An FD to the preference XML file
+         */
+        Settings(int fd);
+
+        /**
+         * @brief Retrieves a particular setting as a string
+         * @param key The key of the setting
+         * @return The string value of the setting
+         */
+        std::string GetString(const std::string &key);
```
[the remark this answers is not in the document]
PixelyIon: Doing this introduces a whole bunch of issues as `std::unordered_map<std::string>::at(std::string_view)` doesn't exist, we'd need to make the entire `unordered_map` into a `std::string_view` one which would require somewhat messy manual hashing. As a result we cannot retrieve the value for `Settings::List` either which isn't optimal, all in all it's much better to leave it as is.

app/src/main/cpp/skyline/common.h
```diff
@@ -104,14 +144,38 @@ namespace skyline {
             return ticks;
         }
 
+        /**
+         * @brief A way to implicitly convert a pointer to uintptr_t and leave it unaffected if it isn't a pointer
+         */
+        template<class T>
```
bylaws, line 150: Don't we use typename?

app/src/main/cpp/skyline/common/circular_queue.h
```diff
@@ -0,0 +1,114 @@
...
+        std::vector<u8> vector; //!< The internal vector holding the circular queue's data, we use a byte vector due to the default item construction/destruction semantics not being appropriate for a circular buffer
+        Type *start{reinterpret_cast<Type *>(vector.begin().base())}; //!< The start/oldest element of the queue
+        Type *end{reinterpret_cast<Type *>(vector.begin().base())}; //!< The end/newest element of the queue
+        std::mutex consumptionMutex;
+        std::condition_variable consumeCondition;
+        std::mutex productionMutex;
+        std::condition_variable produceCondition;
+
+      public:
+        inline CircularQueue(size_t size) : vector((size + 1) * sizeof(Type)) {}
```
bylaws, line 24: add a comment to explain why + 1 is needed

app/src/main/cpp/skyline/common/signal.h
```diff
@@ -0,0 +1,70 @@
...
+namespace skyline::signal {
+    /**
+     * @brief An exception object that is designed specifically to hold Linux signals
+     * @note This doesn't inherit std::exception as it shouldn't be caught as such
+     * @note Refer to the manpage siginfo(3) for information on members
+     */
+    class SignalException {
+      public:
+        int signal{};
+        u64 pc{};
```
bylaws, line 17: uintptr_t

app/src/main/cpp/skyline/kernel/ipc.cpp
```diff
@@ -95,41 +95,41 @@ namespace skyline::kernel::ipc {
         payloadOffset = cmdArg;
 
         if (payload->magic != util::MakeMagic<u32>("SFCI") && (header->type != CommandType::Control && header->type != CommandType::ControlWithContext)) // SFCI is the magic in received IPC messages
-            state.logger->Debug("Unexpected Magic in PayloadHeader: 0x{:X}", u32(payload->magic));
+            state.logger->Debug("Unexpected Magic in PayloadHeader: 0x{:X}", static_cast<u32>(payload->magic));
```
bylaws, line 98: Can we ellide this message in cases where it *is* expected? it is kinda confusing

app/src/main/cpp/skyline/kernel/svc.cpp
```diff
@@ -21,374 +22,466 @@ namespace skyline::kernel::svc {
...
+            size_t size{state.ctx->gpr.x2};
             if (!util::PageAligned(size)) {
-                state.ctx->registers.w0 = result::InvalidSize;
+                state.ctx->gpr.w0 = result::InvalidSize;
                 state.logger->Warn("svcMapSharedMemory: 'size' {}: 0x{:X}", size ? "not page aligned" : "is zero", size);
                 return;
             }
 
-            auto permission{*reinterpret_cast<memory::Permission *>(&state.ctx->registers.w3)};
+            auto permission{*reinterpret_cast<memory::Permission *>(&state.ctx->gpr.w3)};
```
bylaws, line 422: The reinterpret cast can be dropped here

app/src/main/cpp/skyline/kernel/svc.cpp
```diff
@@ -21,374 +22,466 @@ namespace skyline::kernel::svc {
...
+        size_t size{state.ctx->gpr.x2};
         if (!util::PageAligned(size)) {
-            state.ctx->registers.w0 = result::InvalidSize;
+            state.ctx->gpr.w0 = result::InvalidSize;
             state.logger->Warn("svcCreateTransferMemory: 'size' {}: 0x{:X}", size ? "not page aligned" : "is zero", size);
             return;
         }
 
-        auto permission{*reinterpret_cast<memory::Permission *>(&state.ctx->registers.w3)};
+        auto permission{*reinterpret_cast<memory::Permission *>(&state.ctx->gpr.w3)};
```
bylaws, line 455: and here

app/src/main/cpp/skyline/kernel/types/KThread.cpp
```diff
@@ -1,45 +1,169 @@
...
+        "MRS X0, TPIDR_EL0\n\t"
+        "MSR TPIDR_EL0, %x0\n\t" // Set TLS to ThreadContext
+        "STR X0, [%x0, #0x2A0]\n\t" // Write ThreadContext::hostTpidrEl0
+        "MOV X0, SP\n\t"
+        "STR X0, [%x0, #0x2A8]\n\t" // Write ThreadContext::hostSp
+        "MOV SP, %x1\n\t" // Replace SP with guest stack
+        "MOV LR, %x2\n\t" // Store entry in Link Register so it is jumped to on return
+        "MOV X0, %x3\n\t" // Store the argument in X0
+        "MOV X1, %x4\n\t" // Store the thread handle in X1, NCA applications require this
+        "MOV X2, XZR\n\t"
```
bylaws, line 65: add a comment here 'Zero out all other registers'

app/src/main/cpp/skyline/loader/nca.cpp
```diff
@@ -13,35 +12,41 @@ namespace skyline::loader {
...
+        state.process->memory.InitializeVmm(process->npdm.meta.flags.type);
+
         auto loadInfo{NsoLoader::LoadNso(nsoFile, process, state)};
         u64 offset{loadInfo.size};
-        u64 base{loadInfo.base};
+        u8 *base{loadInfo.base};
+        void *entry{loadInfo.entry};
 
-        state.logger->Info("Loaded nso 'rtld' at 0x{:X}", base);
+        state.logger->Info("Loaded nso 'rtld' at 0x{:X} (.text @ 0x{:X})", base, entry);
```
bylaws, line 30: rtld.nso, same print as others

app/src/main/cpp/skyline/nce.cpp
```diff
@@ -12,405 +13,311 @@
...
+        if (*tls) {
+            const auto &state{*reinterpret_cast<ThreadContext *>(*tls)->state};
+            if (signal != SIGINT) {
+                state.logger->Warn("Thread #{} has crashed due to signal: {}", state.thread->id, strsignal(signal));
+
+                std::string raw;
+                std::string trace;
+                std::string cpuContext;
+
+                const auto &ctx{reinterpret_cast<ucontext *>(context)->uc_mcontext};
```
bylaws, line 58: is there a reason for casting ucontext * to ucontext * here?

app/src/main/cpp/skyline/nce/instructions.h
```diff
@@ -3,7 +3,7 @@
 
 #include <common.h>
 
-namespace skyline {
+namespace skyline::nce {
     namespace regs {
         enum X { X0, X1, X2, X3, X4, X5, X6, X7, X8, X9, X10, X11, X12, X13, X14, X15, X16, X17, X18, X19, X20, X21, X22, X23, X24, X25, X26, X27, X28, X29, X30 };
```
bylaws, line 8: :wrench:

app/src/main/cpp/skyline/os.cpp
```diff
@@ -13,58 +12,31 @@
...
 namespace skyline::kernel {
-    OS::OS(std::shared_ptr<JvmManager> &jvmManager, std::shared_ptr<Logger> &logger, std::shared_ptr<Settings> &settings, const std::string &appFilesPath) : state(this, process, jvmManager, settings, logger), memory(state), serviceManager(state), appFilesPath(appFilesPath) {}
+    OS::OS(std::shared_ptr<JvmManager> &jvmManager, std::shared_ptr<Logger> &logger, std::shared_ptr<Settings> &settings, const std::string &appFilesPath) : state(this, jvmManager, settings, logger), serviceManager(state), appFilesPath(appFilesPath) {}
 
     void OS::Execute(int romFd, loader::RomFormat romType) {
         auto romFile{std::make_shared<vfs::OsBacking>(romFd)};
         auto keyStore{std::make_shared<crypto::KeyStore>(appFilesPath)};
 
-        if (romType == loader::RomFormat::NRO) {
+        if (romType == loader::RomFormat::NRO)
```
bylaws, line 21: let's use a switch here

bylaws (changes requested):

app/src/main/cpp/skyline/kernel/svc.cpp
PixelyIon: I found an unrelated issue due to this, I must've assumed that the rwx fields were bitfield booleans (I remember being them like this), it must've changed at some point and broken them. I've now fixed it by making `Permission` a `union` and having an appropriate constructor.

app/src/main/cpp/skyline/nce.cpp
PixelyIon: No, it's a remnant from past code back when that argument was `void *` (That's what it should be due to `sa_action` in the `sigaction` structure but it is always `ucontext *` so I directly made it `ucontext *`)

bylaws:
:partying_face: :champagne: :japanese_castle:

bylaws (approved):

PixelyIon:
The final commit 34a66de6a505dd96507647352519f31848671ec1 has issues which lead to a segmentation fault on any IPC request which has an incorrect magic but isn't a domain request, `SignalException::what` was also broken leading to cascading failure and exiting.

The fixes to these issues have been amended into the merged commit in [master](https://github.com/skyline-emu/skyline/tree/master/) at 8564edcb167abcecba9023c31f010de8c47e1fc8 [[diff](https://github.com/skyline-emu/skyline/compare/34a66de6a505dd96507647352519f31848671ec1..8564edcb167abcecba9023c31f010de8c47e1fc8)]

PR #110, opened 2020-11-21
UI Redesign

Grarak:
- Change primary colors to white/black for light/dark respectively
- Accent color stays red
- Replace refresh icon with pull to refresh
- Replace toolbar with custom search bar
- Change grid design layout
- Basic implementation of ViewModel
- Usage of kotlin coroutines
- Make OSC visible if controller id 0 is enabled


Dyldaboss:
Where do I get the NRO file

FAROK000:
Grarak
Ammn the control buttons are little bit annoying when moving a, b, x, y alone in the screen  the same for d-pad
It must be look like  ppsspp (all move together).
Just an opinion 😊

PixelyIon (approved):
After an extensive review of the code, I've determined yet again that you are better at Kotlin than me and it even being reviewed by me is pointless.

IvarWithoutBones (approved):
The redesign looks beautiful, great job!

ezequielbsb25:
> The redesign looks beautiful, great job!
Indeed, but after the redesign update the Sonic xci was didn't find/recognized by the emulator

PixelyIon:
XCIs do work fine on my end so it could be a local issue, head on over to our [Discord](https://discord.gg/XnbXNQM) so we can resolve the issue.

IvarWithoutBones:
This has been resolved on Discord, @ezequielbsb25 had a dump with a wrong file extension causing the emulator to not detect it.

app/src/main/java/emu/skyline/KeyReader.kt
```diff
@@ -26,60 +26,53 @@ object KeyReader {
...
-                        if (value.length != 32 && !isHexString(value))
-                            return@useLines false
-                    }
-                    KeyType.Prod -> {
-                        if (!key.contains("_"))
-                            return@useLines false
-                        if (!isHexString(value))
-                            return@useLines false
+        tmpOutputFile.bufferedWriter().use { writer ->
+            val valid = inputStream!!.bufferedReader().useLines {
```
sspacelynx, line 38: Is the not-null assertion operator avoidable here? I'd avoid it in all cases when possible, maybe by adding some error handling and displaying a Toast/Snackbar to the user if necessary.

app/src/main/java/emu/skyline/RomProvider.kt
```diff
@@ -0,0 +1,38 @@
...
+                addEntries(fileFormats, file, entries)
+            } else {
+                fileFormats[file.name?.substringAfterLast(".")?.toLowerCase()]?.let { romFormat->
+                    entries.getOrPut(romFormat, { arrayListOf() }).add(RomFile(context, romFormat, file.uri).appEntry)
+                }
+            }
+        }
+    }
+
+    fun loadRoms(searchLocation : Uri) = DocumentFile.fromTreeUri(context, searchLocation)!!.let { documentFile ->
```
sspacelynx, line 33: Same thing about the not-null assertion operator here.

build.gradle
```diff
@@ -1,27 +1,33 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-    ext.kotlin_version = '1.4.30'
+    ext {
+        kotlin_version = '1.4.30'
```
sspacelynx, line 5: Nitpick here, kotlin can be bumped to 1.5 but I assume you were probably going to do that anyway.

sspacelynx (commented):
Code looks awesome, your Kotlin is definitly better than mine too.
I'd avoid the not-null assertion operator at all costs, there's a few that I did not report individually in the controllers' code. They might trigger only in rare situations, but I still think they should be handled appropriately.

app/src/main/java/emu/skyline/KeyReader.kt
Grarak: `inputStream` can only be `null` when the content provider has crashed. This shouldn't happen, unless there is something wrong with the OS

app/src/main/java/emu/skyline/RomProvider.kt
Grarak: `fromTreeUri` can only return `null` when the target API is lower than 21

build.gradle
Grarak: I'll do that later, `1.5` wasn't released when I've started with this PR

PixelyIon (approved):
You've really improved the UI so dramatically that it's up there with the best UIs I've ever seen, genuinely amazing job! :tada:

PR #135, opened 2021-03-05
Refactor VFS with better error checking and fix many codestyle issues in the emulator

bylaws:
See the individual commits for details.

Please rebase and merge.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -182,12 +182,28 @@ We generally support the usage of functional programming and lambda, usage of it
 
 ### References
 For passing any parameter which isn't a primitive prefer to use references/const references to pass them into functions or other places as a copy can be avoided that way.  
-In addition, always use a const reference rather than a normal reference unless the argument needs to be modified in-place as the compiler knows the intent far better in that case.
+In addition, always use a const reference rather than a normal reference unless the argument needs to be modified in-place as the compiler knows the intent far better in that case.  
+Note: In constructors if you are copying to a member variable `std::move` is preferred as it allows copy-elision in some circumstances.
 ```cpp
 void DoSomething(const Class& class, u32 primitive);
-void ClassConstructor(const Class& class) : class(class) {} // Make a copy directly from a `const reference` for class member initialization
+void ClassConstructor(Class class) : class(std::move(class)) {} // Make a copy directly from a `const reference` for class member initialization
 ```
 
+### Member Variable Shadowing
+Shadowing of class member variables should be avoided aside from in constructors when they are exclusively used in the initialisation list.
+To avoid shadowing a prefix of p (parameter) or l (local) should be added to the offending variable:
+
+* Correct
+```cpp
+ClassA(ClassB pClassB) : classB(std::move(pClassB)) {
+	classB->Initialise();
+}
+```
+
+* Incorrect
+```c++
+ClassA(ClassB pClassB) : classB(std::move(pClassB)) {}
+```
 ### Range-based Iterators
 Use C++ range-based iterators for any C++ container iteration unless it can be performed better with functional programming (After C++20 when they are merged with the container). In addition, stick to using references/const references using them
 
--- a/app/src/main/cpp/skyline/kernel/memory.cpp
+++ b/app/src/main/cpp/skyline/kernel/memory.cpp
@@ -61,7 +61,9 @@ namespace skyline::kernel {
         if (!base.address)
             throw exception("Cannot find a suitable carveout for the guest address space");
 
-        mmap(reinterpret_cast<void *>(base.address), base.size, PROT_NONE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
+        auto result{mmap(reinterpret_cast<void *>(base.address), base.size, PROT_NONE, MAP_FIXED | MAP_ANONYMOUS | MAP_PRIVATE, -1, 0)};
+        if (result == MAP_FAILED) [[unlikely]]
+            throw exception("Failed to mmap guest address space: {}", strerror(errno));
 
         chunks = {
             ChunkDescriptor{
--- a/app/src/main/cpp/skyline/kernel/memory.h
+++ b/app/src/main/cpp/skyline/kernel/memory.h
@@ -25,7 +25,7 @@ namespace skyline {
              */
             constexpr Permission(bool read, bool write, bool execute) : r(read), w(write), x(execute) {}
 
-            inline bool operator==(const Permission &rhs) const { return (this->r == rhs.r && this->w == rhs.w && this->x == rhs.x); }
+            inline bool operator==(const Permission &rhs) const { return r == rhs.r && w == rhs.w && x == rhs.x; }
 
             inline bool operator!=(const Permission &rhs) const { return !operator==(rhs); }
 
@@ -176,7 +176,7 @@ namespace skyline {
             constexpr MemoryState KernelStack{0x00002013};
             constexpr MemoryState CodeReadOnly{0x00402214};
             constexpr MemoryState CodeWritable{0x00402015};
-        };
+        }
 
         struct Region {
             u64 address;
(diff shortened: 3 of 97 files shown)


app/src/main/cpp/skyline/kernel/memory.cpp
```diff
@@ -61,7 +61,9 @@ namespace skyline::kernel {
         if (!base.address)
             throw exception("Cannot find a suitable carveout for the guest address space");
 
-        mmap(reinterpret_cast<void *>(base.address), base.size, PROT_NONE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
+        auto ret{mmap(reinterpret_cast<void *>(base.address), base.size, PROT_NONE, MAP_FIXED | MAP_ANONYMOUS | MAP_PRIVATE, -1, 0)};
+        if (ret != reinterpret_cast<void *>(base.address))
```
PixelyIon, line 65:
```suggestion
        auto result{mmap(reinterpret_cast<void *>(base.address), base.size, PROT_NONE, MAP_FIXED | MAP_ANONYMOUS | MAP_PRIVATE, -1, 0)};
        if (result == MAP_FAILED) [[unlikely]]
```
I should mention I will be updating the error if clauses for other system calls with `[[unlikely]]` in another PR

app/src/main/cpp/skyline/kernel/memory.h
```diff
@@ -25,7 +25,7 @@ namespace skyline {
              */
             constexpr Permission(bool read, bool write, bool execute) : r(read), w(write), x(execute) {}
 
-            inline bool operator==(const Permission &rhs) const { return (this->r == rhs.r && this->w == rhs.w && this->x == rhs.x); }
+            inline bool operator==(const Permission &rhs) const { return (r == rhs.r && w == rhs.w && x == rhs.x); }
```
PixelyIon, line 28: Remove the brackets around the entire statement, that's seemingly pointless

app/src/main/cpp/skyline/loader/nro.cpp
```diff
@@ -5,10 +5,11 @@
 #include <kernel/types/KProcess.h>
 #include <vfs/nacp.h>
 #include <vfs/region_backing.h>
+
```
PixelyIon, line 8: There's no need for a space here

app/src/main/cpp/skyline/loader/nso.cpp
```diff
@@ -4,10 +4,11 @@
 #include <lz4.h>
 #include <nce.h>
 #include <kernel/types/KProcess.h>
+
```
PixelyIon, line 7: Same as above, there's no need for a space here

app/src/main/cpp/skyline/nce/instructions.h
```diff
@@ -19,13 +19,8 @@ namespace skyline::nce {
...
             /**
              * @brief Creates a BRK instruction with a specific immediate value, used for generating BRK opcodes
-             * @param value The immediate value of the instruction
              */
-            constexpr Brk(u16 value) {
-                sig0 = 0x0;
-                this->value = value;
-                sig1 = 0x6A1;
-            }
+            constexpr Brk(u16 value) : sig0(0x0), value(value), sig1(0x6A1) {}
```
PixelyIon, line 23: It's very neat that C++20 finally added default initialization for bitfields, this file looks significantly better with it.

CONTRIBUTING.md
```diff
@@ -183,11 +183,24 @@ We generally support the usage of functional programming and lambda, usage of it
...
 In addition, always use a const reference rather than a normal reference unless the argument needs to be modified in-place as the compiler knows the intent far better in that case.
+However in constructors if you are copying to a member variable `std::move` is preferred as it allows copy-elision in some circumstances.
 ```cpp
 void DoSomething(const Class& class, u32 primitive);
-void ClassConstructor(const Class& class) : class(class) {} // Make a copy directly from a `const reference` for class member initialization
+void ClassConstructor(Class class) : class(std::move(class)) {} // Make a copy directly from a `const reference` for class member initialization
 ```
 
+### Member Variable Shadowing
+Shadowing of class member variables should be avoided aside from in constructors when they are exclusively used in the initialisation list.
```
PixelyIon, line 193: We tend to use the American `initialization` rather than `initialisation`, there's an instance of it directly above this

app/src/main/cpp/skyline/services/fssrv/IDirectory.cpp
```diff
@@ -2,6 +2,7 @@
 // Copyright © 2020 Skyline Team and Contributors (https://github.com/skyline-emu/)
 
 #include <kernel/types/KProcess.h>
+
```
PixelyIon, line 5: Same as above, there's no need for spacing here

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.cpp
```diff
@@ -7,10 +7,11 @@
 #include <services/nvdrv/driver.h>
 #include <services/nvdrv/devices/nvmap.h>
 #include <services/common/fence.h>
+
```
PixelyIon, line 10: *Same as above*, no need for spacing here

app/src/main/cpp/skyline/vfs/backing.h
```diff
@@ -42,13 +53,37 @@ namespace skyline::vfs {
...
+                throw exception("Offset cannot be past the end of a backing");
+
+            if ((size - offset) < output.size())
+                throw exception("Trying to read past the end of a backing: 0x{:X}/0x{:X} (Offset: 0x{:X})", output.size(), size, offset);
+
+            if (ReadUnchecked(output, offset) != size)
+                throw exception("Failed to read the requested size from backing");
+
+            return size;
+        };
```
PixelyIon, line 80: I see that you've done the most optimal thing with the API here :stuck_out_tongue:

app/src/main/cpp/skyline/vfs/os_filesystem.cpp
```diff
@@ -5,6 +5,7 @@
 #include <fcntl.h>
 #include <dirent.h>
 #include <unistd.h>
+
```
PixelyIon, line 8: *Saaaame*

PixelyIon (changes requested):

app/src/main/cpp/skyline/vfs/backing.h
bylaws: 🙏

CONTRIBUTING.md
bylaws: Smh you and your inferior spelling

app/src/main/cpp/skyline/kernel/memory.cpp
bylaws: Is there really any point? Like it really isn't going to have any effect on performance aside from making code more cluttered

PixelyIon: It isn't on a small scale but optimizing branches will have an advantage on a larger scale, it is why projects like the Linux kernel or even Atmosphere do them as well. This also includes things such as arranging branches in the order of probabilities, putting an error branch here like this ends up making it the most probable and puts the related assembly right after it while it should really be in the function epilogue.

bylaws: Smh fine

bylaws:
Feedback addressed

PixelyIon (approved):
Good job, no further issues with this PR

PixelyIon:
There was a bug with this PR, a fix has been deployed in #136

PixelyIon:
There was another bug introduced in this PR, a fix has been deployed directly to [`master`](https://github.com/skyline-emu/skyline/tree/master/) as e90dd672af1a3710bc16a773a5a9f800faa088d6

PR #136, opened 2021-03-06
Fix backing read size check

bylaws:
An incorrect check introduced in #135 has broken all VFS backing reads, this PR corrects that check.


PixelyIon (approved):
LGTM! I'll add a comment in #135 mentioning this so that they're linked.

PR #138, opened 2021-03-06
Add NSP ticket extraction support

sspacelynx:


app/src/main/cpp/skyline/crypto/key_store.cpp
```diff
@@ -38,6 +38,11 @@ namespace skyline::crypto {
         titleKeys.emplace(key, valueArray);
     }
 
+    void KeyStore::PopulateTitleKeys(Key128 keyName, Key128 value) {
```
bylaws, line 41: PopulateTitleKey maybe? No point having it plural when it only does one

app/src/main/cpp/skyline/loader/nsp.cpp
```diff
@@ -62,4 +64,18 @@ namespace skyline::loader {
         icon->Read(buffer);
         return buffer;
     }
+
+    void NspLoader::ExtractTickets(const std::shared_ptr<vfs::Directory> &dir, const std::shared_ptr<crypto::KeyStore> &keyStore) {
+        std::vector<vfs::Ticket> tickets;
+        for (const auto &entry : dir->Read()) {
+            if (entry.name.substr(entry.name.find_last_of('.') + 1) == "tik") {
```
bylaws, line 71: Can drop the {} on the if here

app/src/main/cpp/skyline/loader/nsp.cpp
```diff
@@ -62,4 +64,18 @@ namespace skyline::loader {
         icon->Read(buffer);
         return buffer;
     }
+
+    void NspLoader::ExtractTickets(const std::shared_ptr<vfs::Directory> &dir, const std::shared_ptr<crypto::KeyStore> &keyStore) {
```
bylaws, line 68: You can make this a static function in the c++ file itself and remove it from the header which might be cleaner

bylaws (changes requested):

sspacelynx:
Comments addressed, changes have been squashed as there was no point in keeping a commit that failed to build.

app/src/main/cpp/skyline/vfs/ticket.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+namespace skyline::vfs {
+    enum class SignatureType : u32 {
+        Rsa4096Sha1 = 0x010000,
+        Rsa2048Sha1 = 0x010001,
+        EcdsaSha1 = 0x010002,
+        Rsa4096Sha256 = 0x010003,
+        Rsa2048Sha256 = 0x010004,
+        EcdsaSha256 = 0x010005
+    };
```
PixelyIon, line 14: We tend to keep structures and enumerations in the header file, there's few exceptions to this with structures but enumerations should always be in the header with an appropriate comment, which would be the following in this case:
```cpp
/**
  @url https://switchbrew.org/wiki/Ticket#Signature_type
 **/
```
It should also be noted that we tend to have a trailing comma after our enumerations

app/src/main/cpp/skyline/vfs/ticket.h
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright © 2021 Skyline Team and Contributors (https://github.com/skyline-emu/)
+
+#pragma once
+
+#include <vfs/backing.h>
+#include <crypto/key_store.h>
+
+namespace skyline::vfs {
+    struct Ticket {
```
PixelyIon, line 10: There should be a comment here, at the *very least* linking to Switchbrew but I'd have a `@brief` comment explaining what tickets are and their function on a basic level

app/src/main/cpp/skyline/vfs/ticket.h
```diff
@@ -0,0 +1,33 @@
...
+        u8 masterKeyRevision;
+        u8 _pad2_[0xA];
+        u64 ticketId;
+        u64 deviceId;
+        crypto::KeyStore::Key128 rightsId;
+        u32 accountId;
+        u8 _pad3_[0xC];
+
+        Ticket() = default;
+        Ticket(const std::shared_ptr<vfs::Backing> &backing);
```
PixelyIon, line 30: There should be spacing between all function definitions

app/src/main/cpp/skyline/vfs/ticket.h
```diff
@@ -0,0 +1,33 @@
...
+#pragma once
+
+#include <vfs/backing.h>
+#include <crypto/key_store.h>
+
+namespace skyline::vfs {
+    struct Ticket {
+        enum class TitleKeyType : u8 {
+            Common = 0x0,
+            Personal = 0x1
```
PixelyIon, line 13: You should have per-member comments explaining the differences here:
```suggestion
            Common = 0x0,   //!< The title key will be stored as a 16-byte block
            Personal = 0x1, //!< The title key will be stored as a personalized RSA-2048 message
```

app/src/main/cpp/skyline/vfs/ticket.cpp
```diff
@@ -0,0 +1,39 @@
...
+            case SignatureType::Rsa2048Sha1:
+            case SignatureType::Rsa2048Sha256:
+                offset = 0x140;
+                break;
+            case SignatureType::EcdsaSha1:
+            case SignatureType::EcdsaSha256:
+                offset = 0x80;
+                break;
+            default:
+                throw exception("Could not find valid signature type 0x{0:X}", type);
```
PixelyIon, line 34: Is there any reason to use `0x{0:X}` rather than `0x{:X}`? `libfmt` will assume the indices are in an incrementing order when you don't specify one, which you should only do when you print a single argument twice.

PixelyIon (changes requested):
Great job, I just a few minor nits here and there

app/src/main/cpp/skyline/vfs/ticket.cpp
bylaws: I disagree with this, enumerations in c++ makes more sense to me

app/src/main/cpp/skyline/vfs/ticket.cpp
sspacelynx: Oh sure, I didn't really know what to put here, and I forgot to ask you on discord, will change to the one you suggested.

app/src/main/cpp/skyline/loader/nsp.cpp
```diff
@@ -2,13 +2,30 @@
...
+
+        auto dirContent{dir->OpenDirectory("", {false, true})};
+        for (const auto &entry : dirContent->Read()) {
+            if (entry.name.substr(entry.name.find_last_of('.') + 1) == "tik")
+                tickets.emplace_back(dir->OpenFile(entry.name));
+        }
+
+        for (auto ticket : tickets) {
+            auto titleKey{span(ticket.titleKeyBlock).subspan(0, 16).as<crypto::KeyStore::Key128>()};
+            keyStore->PopulateTitleKey(ticket.rightsId, titleKey);
```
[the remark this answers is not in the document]
sspacelynx: don't know what to do here either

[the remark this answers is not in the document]
bylaws: Leave as is

app/src/main/cpp/skyline/vfs/ticket.h
sspacelynx: How does this look
```
/**
  * @brief The Ticket struct allows easy access to ticket files, a format used to store an encrypted title keys
  * @url https://switchbrew.org/wiki/Ticket
  */
```

PixelyIon: LGTM!

bylaws (approved):
Good

sspacelynx:
Good to go

app/src/main/cpp/skyline/crypto/key_store.cpp
```diff
@@ -38,6 +38,14 @@ namespace skyline::crypto {
         titleKeys.emplace(key, valueArray);
     }
 
+    /**
+     * @note Any title keys which are already in the store will not have their values updated
+     */
```
PixelyIon, line 43: This comment would be in the header rather than the source file

sspacelynx: 🤦‍♂️

FAROK000:
So we don't need the ui of " import title keys" ?

sspacelynx:
Yes, but only for games that contain valid tickets.

PR #140, opened 2021-03-06
Entirely reimplement time services

bylaws:
See the individual commits for detailed descriptions, please rebase and merge.

It should be noted that tzcode currently points to my github, once the code is reviewed we will move it to the skyline org.


app/src/main/cpp/skyline/services/timesrv/time_shared_memory.cpp
```diff
@@ -0,0 +1,144 @@
...
+            u32 updateCount;
+            std::array<u8, 2> enabled;
+        } standardUserSystemClockAutomaticCorrectionEnabledEntry;
+    };
+    static_assert(offsetof(TimeSharedMemoryLayout, localSystemClockContextEntry) == 0x38);
+    static_assert(offsetof(TimeSharedMemoryLayout, networkSystemClockContextEntry) == 0x80);
+    static_assert(offsetof(TimeSharedMemoryLayout, standardUserSystemClockAutomaticCorrectionEnabledEntry) == 0xC8);
+
+    /**
+     * @brief Time Shared Memory uses a double buffered format that alternates writes context data, this is a helper to simplify that
```
PixelyIon, line 29: There should a clear explanation regarding why this function requires memory stores to be strongly ordered if not in the function comment in the function itself (`// The item value must be updated prior to updateCount to prevent reading in an invalid item value`), also fix `that alternates writes context data`.

app/src/main/cpp/skyline/services/timesrv/time_shared_memory.cpp
```diff
@@ -0,0 +1,144 @@
...
+     */
+    template<typename T>
+    static void UpdateTimeSharedMemoryItem(u32 &updateCount, std::array<T, 2> &item, const T &newValue) {
+        u32 newCount{updateCount + 1};
+        item[newCount & 1] = newValue;
+        asm volatile("DMB ISHST"); // 0xA
```
PixelyIon, line 35: It might be better to use [**std::atomic_thread_fence**](https://en.cppreference.com/w/cpp/atomic/atomic_thread_fence) with an appropriate [**std::memory_order**](https://en.cppreference.com/w/cpp/atomic/memory_order)

app/src/main/cpp/skyline/services/timesrv/time_shared_memory.cpp
```diff
@@ -0,0 +1,144 @@
...
+     */
+    template<typename T>
+    static T ReadTimeSharedMemoryItem(u32 &updateCount, std::array<T, 2> &item) {
+        u32 checkUpdateCount{};
+        T out{};
+
+        do {
+            checkUpdateCount = updateCount;
+            out = item[updateCount & 1];
+            asm volatile("DMB ISHLD"); // 0x9
```
PixelyIon, line 50: Same as above, inline ASM should be the last resort

app/src/main/cpp/skyline/services/timesrv/time_shared_memory.h
```diff
@@ -0,0 +1,110 @@
...
+#pragma once
+
+#include <common.h>
+#include <common/uuid.h>
+#include <kernel/types/KEvent.h>
+#include <kernel/types/KSharedMemory.h>
+#include <services/timesrv/common.h>
+
+namespace skyline::service::timesrv::core {
+    struct TimeSharedMemoryLayout;
```
PixelyIon, line 13: As determined by our (somewhat informal) convention, if there's any links to a structure from a header then that structure should be placed in the header as well

app/src/main/cpp/skyline/services/timesrv/time_shared_memory.h
```diff
@@ -0,0 +1,110 @@
...
+        void SetStandardUserSystemClockAutomaticCorrectionEnabled(bool enabled);
+    };
+
+    /**
+     * @brief Base class for callbacks that run after a system clock context is updated
+     */
+    class SystemClockContextUpdateCallback {
+      private:
+        std::list<std::shared_ptr<kernel::type::KEvent>> operationEventList; //!< List of KEvents to be signalled when this callback is called
+        std::mutex mutex; //!< Protects access to operationEventList
```
PixelyIon, line 50: Synchronizes >>>>>>>>>>> Protects

app/src/main/cpp/skyline/services/timesrv/time_shared_memory.h
```diff
@@ -0,0 +1,110 @@
...
+
+        void SetStandardUserSystemClockAutomaticCorrectionEnabled(bool enabled);
+    };
+
+    /**
+     * @brief Base class for callbacks that run after a system clock context is updated
+     */
+    class SystemClockContextUpdateCallback {
+      private:
+        std::list<std::shared_ptr<kernel::type::KEvent>> operationEventList; //!< List of KEvents to be signalled when this callback is called
```
PixelyIon, line 49: Same as above with just pluralizing it rather than tagging along `List` as a prefix, there might've been instances of this I've missed so do check for them

app/src/main/cpp/skyline/services/timesrv/timezone_manager.cpp
```diff
@@ -0,0 +1,130 @@
...
+        memcpy(ruleOut.data(), ruleObj, ruleOut.size_bytes());
+        tz_tzfree(ruleObj);
+        return {};
+    }
+
+    ResultValue<FullCalendarTime> TimeZoneManager::ToCalendarTime(tz_timezone_t pRule, PosixTime posixTime) {
+        struct tm tmp{};
+        auto posixCalendarTime{tz_localtime_rz(pRule, &posixTime, &tmp)};
+        if (!posixCalendarTime)
+            return result::PermissionDenied; // Not the proper error here but *fine*
```
PixelyIon, line 95: Is that Markdown I see? :roll_eyes:

app/src/main/cpp/skyline/services/timesrv/timezone_manager.cpp
```diff
@@ -0,0 +1,130 @@
...
+            },
+            .additionalInfo{
+                .dayOfWeek = static_cast<u32>(posixCalendarTime->tm_wday),
+                .dayOfYear = static_cast<u32>(posixCalendarTime->tm_yday),
+                .dst = static_cast<u32>(posixCalendarTime->tm_isdst),
+                .gmtOffset = static_cast<i32>(posixCalendarTime->tm_gmtoff),
+            },
+        };
+
+        std::string_view(posixCalendarTime->tm_zone).copy(out.additionalInfo.timeZoneName.data(), strlen(posixCalendarTime->tm_zone));
```
PixelyIon, line 114: Assign the string view as a variable since using it in a `copy` here with `strlen` is doing the same operation twice, use `string_view::size` instead

app/src/main/cpp/skyline/vfs/android_asset_backing.cpp
```diff
@@ -0,0 +1,30 @@
...
+    AndroidAssetBacking::~AndroidAssetBacking() {
+        AAsset_close(backing);
+    }
+
+    size_t AndroidAssetBacking::ReadImpl(span<u8> output, size_t offset) {
+        if (AAsset_seek64(backing, offset, SEEK_SET) != offset)
+            throw exception("Failed to seek asset position");
+
+        auto ret{AAsset_read(backing, output.data(), output.size())};
+        if (ret < 0)
```
PixelyIon, line 25: *throws a fit about the lack of `[[unlikely]]` for those sweet sweet micro-optimization points*

app/src/main/cpp/skyline/vfs/android_asset_backing.h
```diff
@@ -0,0 +1,28 @@
...
+
+namespace skyline::vfs {
+    /**
+     * @brief The AndroidAssetBacking class provides the backing abstractions for the AAsset Android API
+     * @note This is NOT thread safe NOR should it be shared across threads.
+     * @note This will take ownership of the backing asset passed into it
+     */
+    class AndroidAssetBacking : public Backing {
+      private:
+        AAsset *backing; //!< The backing AAsset object we abstract
```
PixelyIon, line 18: I recommend renaming this to prevent confusion with the backing abstraction in itself, it took me a minute to realize what this was
```suggestion
        AAsset *asset; //!< The underlying NDK AAsset object
```

app/src/main/cpp/skyline/vfs/android_asset_backing.h
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright © 2021 Skyline Team and Contributors (https://github.com/skyline-emu/)
+
+#pragma once
+
+#include "backing.h"
+
+struct AAsset;
```
PixelyIon, line 8:
```suggestion
struct AAsset; //!< https://developer.android.com/ndk/reference/group/asset#aasset
```

app/src/main/cpp/skyline/vfs/android_asset_filesystem.cpp
```diff
@@ -0,0 +1,35 @@
...
+        auto file{AAssetManager_open(backing, path.c_str(), AASSET_MODE_RANDOM)};
+        if (file == nullptr)
+            return nullptr;
+
+        return std::make_shared<AndroidAssetBacking>(file, mode);
+    }
+
+    std::optional<Directory::EntryType> AndroidAssetFileSystem::GetEntryTypeImpl(const std::string &path) {
+        // Tries to open as a file then as a directory in order to check the type
+        // This is really hacky but it at least it works
```
PixelyIon, line 21:
```suggestion
        // It's really hacky but at least it works
```

app/src/main/cpp/skyline/vfs/android_asset_filesystem.cpp
```diff
@@ -0,0 +1,35 @@
...
+        if (AAssetManager_open(backing, path.c_str(), AASSET_MODE_RANDOM) != nullptr)
+            return Directory::EntryType::File;
+
+        if (AAssetManager_openDir(backing, path.c_str()) != nullptr)
+            return Directory::EntryType::Directory;
+
+        // Doesn't exit
```
PixelyIon, line 28:
```suggestion
        // The path doesn't exist
```

app/src/main/cpp/skyline/vfs/android_asset_filesystem.h
```diff
@@ -0,0 +1,28 @@
...
+
+struct AAssetManager;
+
+namespace skyline::vfs {
+    /**
+     * @brief  The AndroidAssetFileSystem class provides the filesystem backing abstractions for the AAsset Android API
+     */
+    class AndroidAssetFileSystem : public FileSystem {
+      private:
+        AAssetManager *backing; //!< The backing manager of the filesystem
```
PixelyIon, line 16: Same as above, naming this `backing` rather than something like `assetManager` is very confusing as it can be conflated with our VFS backing abstraction

PixelyIon (changes requested):

app/src/main/cpp/skyline/common.h
```diff
@@ -71,15 +71,53 @@ namespace skyline {
...
+        std::optional<T> value;
+
+      public:
+        Result result;
+
+        constexpr ResultValue(T value) : value(value) {};
+
+        constexpr ResultValue(Result result) : result(result) {};
+
+        template<typename G>
```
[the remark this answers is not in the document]
bylaws: Just a name? T was taken

app/src/main/cpp/skyline/common/uuid.cpp
```diff
@@ -0,0 +1,76 @@
...
+                        u8 reserved : 2;
+                    };
+                };
+                u8 clockSeqLow;
+
+                union __attribute__((packed)) {
+                    u64 node : 48;
+
+                    struct {
+                        std::array<u8, 6> nodeRaw;
```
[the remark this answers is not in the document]
bylaws: Sure

app/src/main/cpp/skyline/services/glue/ITimeZoneService.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+        auto locationName{span(request.Pop<timesrv::LocationName>()).as_string(true)};
+
+        // TODO Checked
+        auto timeZoneBinaryFile{state.os->assetFileSystem->OpenFile(fmt::format("tzdata/zoneinfo/{}", locationName))};
+        std::vector<u8> timeZoneBinaryBuffer(timeZoneBinaryFile->size);
+        timeZoneBinaryFile->Read(timeZoneBinaryBuffer);
+        auto ret{core->SetDeviceLocationNameWithTimeZoneBinary(span(locationName).as_string(true), timeZoneBinaryBuffer)};
+        if (ret)
+            return ret;
```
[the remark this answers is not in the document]
bylaws: Disagree, Ret is very common in other projects and easier to type

app/src/main/cpp/skyline/services/glue/ITimeZoneService.cpp
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    Result ITimeZoneService::LoadTimeZoneRule(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
+        auto locationName{span(request.Pop<timesrv::LocationName>()).as_string(true)};
+
+        // TODO Checked
+        auto timeZoneBinaryFile{state.os->assetFileSystem->OpenFile(fmt::format("tzdata/zoneinfo/{}", locationName))};
+        std::vector<u8> timeZoneBinaryBuffer(timeZoneBinaryFile->size);
+        timeZoneBinaryFile->Read(timeZoneBinaryBuffer);
+        return core->ParseTimeZoneBinary(timeZoneBinaryBuffer, request.outputBuf.at(0));
```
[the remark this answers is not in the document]
bylaws: Don't think so, better to throw on it anyhow as it would suggest a real bug deeper down (eg game reading invalid timezone)

app/src/main/cpp/skyline/services/glue/ITimeZoneService.cpp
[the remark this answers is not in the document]
PixelyIon: The issue is moreso that `ret` is stupid as it doesn't at all imply what the contents of it are which is what variable names are for rather than what the variable will be used for

app/src/main/cpp/skyline/services/timesrv/IStaticService.cpp
```diff
@@ -1,36 +1,239 @@
...
+
+        out.userCalendarTime = userCalendarTime->calendarTime;
+        out.userCalendarAdditionalInfo = userCalendarTime->additionalInfo;
+
+        // Not necessarily a fatal error if this fails
+        auto networkPosixTime{ClockSnapshot::GetCurrentTime(out.steadyClockTimePoint, out.networkContext)};
+        if (networkPosixTime)
+            out.networkPosixTime = *networkPosixTime;
+        else
+            out.networkPosixTime = 0;
```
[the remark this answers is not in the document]
bylaws: Eh, I suppose not but think it's cleaner personally , otherwise I can just add a comment about it I suppose

app/src/main/cpp/skyline/common.h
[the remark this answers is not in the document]
PixelyIon: Template names are supposed to make some amount of sense, this level of arbitrary doesn't feel right for naming. We also tend to not use `T` and use `Type` instead for template types.

app/src/main/cpp/skyline/services/timesrv/ISystemClock.h
```diff
@@ -3,41 +3,45 @@
...
-     * @brief ISystemClock is used to retrieve and set time
+     * @brief ISystemClock is used to interface with timesrv system clocks
      * @url https://switchbrew.org/wiki/PSC_services#ISystemClock
      */
     class ISystemClock : public BaseService {
-      public:
-        const SystemClockType type;
+      private:
+        core::SystemClockCore &core;
+        bool writeClock;
```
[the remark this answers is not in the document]
bylaws: Yes

app/src/main/cpp/skyline/services/timesrv/ITimeZoneService.cpp
```diff
@@ -1,32 +1,132 @@
...
+
+        if (calendarTime)
+            response.Push(*calendarTime);
+
+        return calendarTime;
+    }
 
     Result ITimeZoneService::ToCalendarTimeWithMyRule(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
-        auto &time{request.Pop<u64>()};
-        auto calender{*std::gmtime(reinterpret_cast<const time_t *>(&time))};
```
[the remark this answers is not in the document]
bylaws: 😡

app/src/main/cpp/skyline/services/timesrv/core.h
```diff
@@ -0,0 +1,323 @@
...
+        /**
+         * @brief The internal offset is the offset between the raw steady clock time and the target time of this steady clock
+         */
+        virtual TimeSpanType GetInternalOffset() {
+            return {};
+        }
+
+        virtual void SetInternalOffset(TimeSpanType offset) {
+            return;
+        }
```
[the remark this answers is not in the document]
bylaws: Nope, this is different, Nintendo does this too, only the Standard clock has an internal offset. The note is described in a more appropriate place anyhow already

app/src/main/cpp/skyline/services/timesrv/ITimeZoneService.cpp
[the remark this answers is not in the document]
PixelyIon: I understand your jealousy of my innovative ways..

app/src/main/cpp/skyline/services/timesrv/time_shared_memory.cpp
```diff
@@ -0,0 +1,144 @@
...
+        ClockContextEntry<SystemClockContext> localSystemClockContextEntry;
+        ClockContextEntry<SystemClockContext> networkSystemClockContextEntry;
+        struct __attribute__((packed)) {
+            u32 updateCount;
+            std::array<u8, 2> enabled;
+        } standardUserSystemClockAutomaticCorrectionEnabledEntry;
+    };
+    static_assert(offsetof(TimeSharedMemoryLayout, localSystemClockContextEntry) == 0x38);
+    static_assert(offsetof(TimeSharedMemoryLayout, networkSystemClockContextEntry) == 0x80);
+    static_assert(offsetof(TimeSharedMemoryLayout, standardUserSystemClockAutomaticCorrectionEnabledEntry) == 0xC8);
```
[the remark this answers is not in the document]
bylaws: Size doesn't exactly matter as there is nothing preventing it being larger eg in an update, I suppose I can assert that it's less than shmem size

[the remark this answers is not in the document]
PixelyIon: I agree, size doesn't matter

app/src/main/cpp/skyline/common.h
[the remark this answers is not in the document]
bylaws: Hm? I swear we use T in many places

app/src/main/cpp/skyline/services/glue/ITimeZoneService.cpp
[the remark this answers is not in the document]
bylaws: Neither does result?

app/src/main/cpp/skyline/services/timesrv/time_shared_memory.h
bylaws: Hm? Disagree, placing it in the c++ file makes much more sense here

app/src/main/cpp/skyline/services/glue/ITimeZoneService.cpp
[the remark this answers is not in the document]
PixelyIon: Isn't the return value's type `Result` and therefore contains a `result` value?

app/src/main/cpp/skyline/vfs/android_asset_filesystem.cpp
bylaws: This is makes more sense to me

app/src/main/cpp/skyline/services/glue/ITimeZoneService.cpp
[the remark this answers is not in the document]
bylaws: Yes, just the same as ret contains a return value

app/src/main/cpp/skyline/vfs/android_asset_filesystem.cpp
PixelyIon: I'm not too fussed about that, just remove the `it` after the `but`

app/src/main/cpp/skyline/services/glue/ITimeZoneService.cpp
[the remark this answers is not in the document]
PixelyIon: No, `ret` describes what's done with a variable rather than it's contents. The variable's contents are not `return` but a result instead.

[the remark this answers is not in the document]
bylaws: Smh you're wrong but I concede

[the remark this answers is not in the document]
PixelyIon: Music to my ears :relaxed:

app/src/main/cpp/skyline/services/timesrv/time_shared_memory.h
bylaws: Especially considering the locality to accessor functions etc, if you take this as a rule you would need GlobalServiceState to be in header (which wouldn't be very good at all)

PixelyIon: "Smh you're wrong but I concede"

app/src/main/cpp/skyline/services/timesrv/core.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+                if (std::next(it) != buffer.end())
+                    prev = std::next(it);
+            }
+        }
+
+        // TODO Checked
+        auto timeZoneBinaryVersionFile{state.os->assetFileSystem->OpenFile("tzdata/version.txt")};
+        std::array<u8, 0x10> timeZoneBinaryVersion{};
+        timeZoneBinaryVersionFile->Read(timeZoneBinaryVersion);
```
PixelyIon, line 290: `tzdata/version.txt` is not `0x10` bytes long which'll lead to `Backing::Read` throwing, you should've tested `time` after rebasing it on `vfs-codestyle`.

PixelyIon (changes requested):

app/src/main/cpp/skyline/services/timesrv/core.h
```diff
@@ -0,0 +1,323 @@
...
+
+        StandardSteadyClockCore standardSteadyClock;
+        TickBasedSteadyClockCore tickBasedSteadyClock;
+        StandardLocalSystemClockCore localSystemClock;
+        StandardNetworkSystemClockCore networkSystemClock;
+        StandardUserSystemClockCore userSystemClock;
+        EphemeralNetworkSystemClockCore empheralSystemClock;
+
+        TimeZoneManager timeZoneManager;
+        std::vector<LocationName> locationNameList; //!< N stores in glue but we are fine putting it here
```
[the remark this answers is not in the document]
bylaws: N uses locationNameList so makes sense to use, will do comment change though

app/src/main/cpp/skyline/services/timesrv/time_shared_memory.cpp
bylaws: None of this generates the same ASM

bylaws (commented):
addressed

bylaws:
and tested, I should add

app/src/main/cpp/skyline/vfs/android_asset_filesystem.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+namespace skyline::vfs {
+    AndroidAssetFileSystem::AndroidAssetFileSystem(AAssetManager *assetManager) : FileSystem(), assetManager(assetManager) {}
+
+    std::shared_ptr<Backing> AndroidAssetFileSystem::OpenFileImpl(const std::string &path, Backing::Mode mode) {
+        auto file{AAssetManager_open(assetManager, path.c_str(), AASSET_MODE_RANDOM)};
+        if (file == nullptr)
+            return nullptr;
+
+        return std::make_shared<AndroidAssetBacking>(file, mode);
```
PixelyIon, line 16: `return file ? std::make_shared<AndroidAssetBacking>(file, mode) : nullptr;`

PixelyIon (changes requested):

bylaws:
Addressed, RTM

PixelyIon (approved):
LGTM, great job! :tada:

PR #172, opened 2021-05-11
Y

sanjay9867:


MichiHD02:
why did he do that

MichiHD02:
ok i'm really curious if you can get the emulator to work and i hope that this project doesn't stop there is already too much work involved

PR #191, opened 2021-06-11
HwOpus decoder service implementation

sspacelynx:


app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
```diff
@@ -0,0 +1,126 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright © 2021 Skyline Team and Contributors (https://github.com/skyline-emu/)
+
+#include "IHardwareOpusDecoder.h"
+#include "results.h"
+
+namespace skyline::service::codec {
+
+    size_t CalculateOutBufferSize(i32 sampleRate, i32 channelCount, i32 frameSize) {
```
bylaws, line 9:
```suggestion
    static size_t CalculateOutBufferSize(i32 sampleRate, i32 channelCount, i32 frameSize) {
```

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
```diff
@@ -0,0 +1,126 @@
...
+        auto pcmDataOut{request.outputBuf.at(0).cast<opus_int16>()};
+
+        auto decodeResult{Decode(opusDataIn, pcmDataOut)};
+        if (!decodeResult) {
+            throw OpusException(decodeResult.result);
+        }
+
+        response.Push(decodeResult->decodedDataSize);
+        response.Push(decodeResult->decodedSampleCount);
+        response.Push<u64>(0);
```
bylaws, line 49: this is a microsecond timer for how long the decode took, you can use chrono for that

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
```diff
@@ -0,0 +1,126 @@
...
+        response.Push(decodeResult->decodedSampleCount);
+        return {};
+    }
+
+    Result IHardwareOpusDecoder::DecodeInterleavedWithPerfOld(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
+        auto opusDataIn{request.inputBuf.at(0)};
+        auto pcmDataOut{request.outputBuf.at(0).cast<opus_int16>()};
+
+        auto decodeResult{Decode(opusDataIn, pcmDataOut)};
+        if (!decodeResult) {
```
bylaws, line 43: no `{` `}`

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
```diff
@@ -0,0 +1,126 @@
...
+        decoderState = opus_decoder_create(sampleRate, channelCount, &result);
+        if (result != OPUS_OK)
+            throw OpusException(result);
+    }
+
+    IHardwareOpusDecoder::~IHardwareOpusDecoder() {
+        opus_decoder_destroy(decoderState);
+    }
+
+    Result IHardwareOpusDecoder::DecodeInterleavedOld(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
```
bylaws, line 24: Make this DecodeInterleavedImpl and have it take a reset boolean and a `reportPerformanceInfo` boolean, that way you avoid needing every func to do the same ipc stuff

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
```diff
@@ -0,0 +1,126 @@
...
+        response.Push<u64>(0);
+        return {};
+    }
+
+    void IHardwareOpusDecoder::ResetContext() {
+        opus_decoder_ctl(decoderState, OPUS_RESET_STATE);
+    }
+
+    ResultValue<IHardwareOpusDecoder::DecodeResult> IHardwareOpusDecoder::Decode(span<u8> dataIn, span<opus_int16> dataOut) {
+        if (dataIn.size() <= sizeof(OpusDataHeader)) {
```
bylaws, line 81: drop `{` `}`

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
```diff
@@ -0,0 +1,126 @@
...
+            return result::BadOpusDataHeader;
+        }
+
+        auto frameSize{util::SwapEndianness(dataIn.as<OpusDataHeader>().size)};
+        auto requiredOutSize{frameSize + sizeof(OpusDataHeader)};
+        if (frameSize > decoderOutputBufferSize || dataOut.size() < requiredOutSize) {
+            return result::OpusBufferTooSmall;
+        }
+
+        dataIn = dataIn.subspan(sizeof(OpusDataHeader));
```
bylaws, line 91: make this a new variable 'sampleDataIn' for something otherwise it's a tad confusing

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    ResultValue<IHardwareOpusDecoder::DecodeResult> IHardwareOpusDecoder::Decode(span<u8> dataIn, span<opus_int16> dataOut) {
+        if (dataIn.size() <= sizeof(OpusDataHeader)) {
+            return result::BadOpusDataHeader;
+        }
+
+        auto frameSize{util::SwapEndianness(dataIn.as<OpusDataHeader>().size)};
+        auto requiredOutSize{frameSize + sizeof(OpusDataHeader)};
+        if (frameSize > decoderOutputBufferSize || dataOut.size() < requiredOutSize) {
```
bylaws, line 87:
```suggestion
        if (frameSize > MaxInputBufferSize || dataIn.size() < requiredInSize)
```

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
```diff
@@ -0,0 +1,126 @@
...
+        opus_decoder_ctl(decoderState, OPUS_RESET_STATE);
+    }
+
+    ResultValue<IHardwareOpusDecoder::DecodeResult> IHardwareOpusDecoder::Decode(span<u8> dataIn, span<opus_int16> dataOut) {
+        if (dataIn.size() <= sizeof(OpusDataHeader)) {
+            return result::BadOpusDataHeader;
+        }
+
+        auto frameSize{util::SwapEndianness(dataIn.as<OpusDataHeader>().size)};
+        auto requiredOutSize{frameSize + sizeof(OpusDataHeader)};
```
bylaws, line 86:
```suggestion
        auto requiredInSize{frameSize + sizeof(OpusDataHeader)};
```

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
```diff
@@ -0,0 +1,126 @@
...
+        if (frameSize > decoderOutputBufferSize || dataOut.size() < requiredOutSize) {
+            return result::OpusBufferTooSmall;
+        }
+
+        dataIn = dataIn.subspan(sizeof(OpusDataHeader));
+        int decodedCount{opus_decode(decoderState, dataIn.data(), frameSize, dataOut.data(), decoderOutputBufferSize, false)};
+
+        if (decodedCount < 0) {
+            Result nnResult{};
```
bylaws, line 95:
```suggestion
```

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
```diff
@@ -0,0 +1,126 @@
...
+            switch (decodedCount) {
+                case OPUS_BAD_ARG:
+                    nnResult = result::OpusBadArgument;
```
bylaws, line 98:
```suggestion
                    return result::OpusBadArgument;
```

do for others

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
bylaws, line 86: don't use auto here

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
```diff
@@ -0,0 +1,126 @@
...
+    void IHardwareOpusDecoder::ResetContext() {
+        opus_decoder_ctl(decoderState, OPUS_RESET_STATE);
+    }
+
+    ResultValue<IHardwareOpusDecoder::DecodeResult> IHardwareOpusDecoder::Decode(span<u8> dataIn, span<opus_int16> dataOut) {
+        if (dataIn.size() <= sizeof(OpusDataHeader)) {
+            return result::BadOpusDataHeader;
+        }
+
+        auto frameSize{util::SwapEndianness(dataIn.as<OpusDataHeader>().size)};
```
bylaws, line 85: don't use auto

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
```diff
@@ -0,0 +1,89 @@
...
+    };
+    static_assert(sizeof(OpusDataHeader) == 0x8);
+
+    size_t CalculateOutBufferSize(i32 sampleRate, i32 channelCount, i32 frameSize);
+
+    static constexpr i32 OpusFullbandSampleRate{48000};
+    static constexpr i32 MaxFrameSizeNormal = OpusFullbandSampleRate * 0.040; //!< 40ms frame size limit for normal decoders
+    static constexpr i32 MaxFrameSizeMultiStream = OpusFullbandSampleRate * 0.120; //!< 120ms frame size limit for multistream decoders
+    static constexpr i32 MaxInputBufferSize{0x600}; //!< This has been REd
+
```
bylaws, line 29: comment needed

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
```diff
@@ -0,0 +1,89 @@
...
+        u32 size; //!< Size of the packet following this header
+        u32 finalRange; //!< Indicates the final range of the codec encoder's entropy coder, this can be left at zero
+    };
+    static_assert(sizeof(OpusDataHeader) == 0x8);
+
+    size_t CalculateOutBufferSize(i32 sampleRate, i32 channelCount, i32 frameSize);
+
+    static constexpr i32 OpusFullbandSampleRate{48000};
+    static constexpr i32 MaxFrameSizeNormal = OpusFullbandSampleRate * 0.040; //!< 40ms frame size limit for normal decoders
+    static constexpr i32 MaxFrameSizeMultiStream = OpusFullbandSampleRate * 0.120; //!< 120ms frame size limit for multistream decoders
```
bylaws, line 27: this is wrong, it's not multistream that has the limit, it's Ex decoders which were added in 12.0.0

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoderManager.h
```diff
@@ -0,0 +1,57 @@
...
+    struct MultiStreamParameters {
+        i32 sampleRate;
+        i32 channelCount;
+        i32 streams; //!< number of streams
+        i32 coupledStreams; //!< Number of stereo streams
+        std::array<u8, 0x100> mappings; //!< Array of channel mappings
+    };
+    static_assert(sizeof(MultiStreamParameters) == 0x110);
+
+    class IHardwareOpusDecoderManager : public BaseService {
```
bylaws, line 22: comment

bylaws (changes requested):

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
sspacelynx: just a heads up, that switch case is currently useless as for every result returned it throws OpusException no matter what

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
sspacelynx: this is also used in `IHardwareOpusDecoderManager.cpp` so cannot be made static

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
sspacelynx: I'm not sure what you are checking here, this is returning `result::OpusBufferTooSmall`

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
sspacelynx: any suggestions? don't know what to put

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
sspacelynx: added a comment instead

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
sspacelynx: I made it actually useful and now it returns result values to guest

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
sspacelynx: what would be a fitting description?

bylaws: MaxFrameSizeMultiEx `//!< 120ms frame size limit for ex decoders added in 12.0.0`

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
bylaws: maximum allocated size of the input buffer. I also should note that you will needa comment for every class

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
bylaws: I think a new variable here is a better idea

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
bylaws: ah fair enough then

app/src/main/cpp/skyline/common.h
```diff
@@ -125,6 +125,7 @@ namespace skyline {
         constexpr u16 DockedResolutionW{1920}; //!< The width component of the docked resolution
         constexpr u16 DockedResolutionH{1080}; //!< The height component of the docked resolution
         // Time
+        constexpr u64 NsInMicrosecond{1000}; //!< The amount of nanosecond in a microsecond
```
bylaws, line 128:
```suggestion
        constexpr u64 NsInMicrosecond{1000}; //!< The amount of nanoseconds in a microsecond
```

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
```diff
@@ -0,0 +1,97 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright © 2021 Skyline Team and Contributors (https://github.com/skyline-emu/)
+
+#include "IHardwareOpusDecoder.h"
+#include "results.h"
+#include <services/timesrv/common.h>
+
+namespace skyline::service::codec {
+
```
bylaws, line 9: <snip>

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
```diff
@@ -0,0 +1,97 @@
...
+        i32 requiredInSize{static_cast<i32>(opusPacketSize + sizeof(OpusDataHeader))};
+        if (opusPacketSize > MaxInputBufferSize || dataIn.size() < requiredInSize)
+            return result::OpusBufferTooSmall;
+
+        // Subtract header from the input buffer to get the opus packet
+        dataIn = dataIn.subspan(sizeof(OpusDataHeader));
+
+        auto perfTimer = timesrv::TimeSpanType::FromNanoseconds(util::GetTimeNs());
+        i32 decodedCount{opus_decode(decoderState, dataIn.data(), opusPacketSize, dataOut.data(), decoderOutputBufferSize, false)};
+        perfTimer = util::GetTimeNs() - perfTimer;
```
bylaws, line 66:
```suggestion
        perfTimer = timesrv::TimeSpanType::FromNanoseconds(util::GetTimeNs()) - perfTimer;
```

app/src/main/cpp/skyline/services/codec/results.h
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright © 2020 Skyline Team and Contributors (https://github.com/skyline-emu/)
```
bylaws, line 2: I agree

bylaws (changes requested):

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
sspacelynx: oh right thank you I missed this

app/src/main/cpp/skyline/services/codec/results.h
PixelyIon: I disagree, can we move to another license?

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
```diff
@@ -0,0 +1,96 @@
...
+            return result::BadOpusDataHeader;
+
+        opus_int32 opusPacketSize{static_cast<opus_int32>(util::SwapEndianness(dataIn.as<OpusDataHeader>().size))};
+        i32 requiredInSize{static_cast<i32>(opusPacketSize + sizeof(OpusDataHeader))};
+        if (opusPacketSize > MaxInputBufferSize || dataIn.size() < requiredInSize)
+            return result::OpusBufferTooSmall;
+
+        // Subtract header from the input buffer to get the opus packet
+        auto sampleDataIn = dataIn.subspan(sizeof(OpusDataHeader));
+
```
bylaws, line 62: Use {} for initialising all variables

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
```diff
@@ -0,0 +1,145 @@
...
+    static constexpr i32 MaxInputBufferSize{0x600}; //!< Maximum allocated size of the input buffer
+
+    /**
+     * @brief This service decodes an Opus audio stream. Always created with OpenHardwareOpusDecoder by IHardwareOpusDecoderManager
+     * @url https://switchbrew.org/wiki/Audio_services#IHardwareOpusDecoder
+     */
+    class IHardwareOpusDecoder : public BaseService {
+      private:
+        OpusDecoder *decoderState;
+        const i32 sampleRate;
```
bylaws, line 37: We don't use const

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
```diff
@@ -0,0 +1,96 @@
...
+                case OPUS_BUFFER_TOO_SMALL:
+                    return result::OpusBufferTooSmall;
+                case OPUS_INTERNAL_ERROR:
+                    return result::OpusInternalError;
+                case OPUS_INVALID_PACKET:
+                    return result::OpusInvalidPacket;
+                case OPUS_INVALID_STATE:
+                    return result::OpusInvalidState;
+                case OPUS_UNIMPLEMENTED:
+                    return result::OpusUnimplemented;
```
bylaws, line 80: These results should never be returned to hos, all should throw an exception containing the result value

bylaws: Also add a comment detailing wjy

bylaws (changes requested):

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
sspacelynx: Should I get rid of that switch case and just throw OpusException? It does handle error codes and error strings by itself so that switch case would be redundant

bylaws: Uh you also need to catch the invalid size results as well but sure as long as you do that

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
```diff
@@ -0,0 +1,78 @@
...
+    void IHardwareOpusDecoder::ResetContext() {
+        opus_decoder_ctl(decoderState, OPUS_RESET_STATE);
+    }
+
+    Result IHardwareOpusDecoder::DecodeInterleavedImpl(ipc::IpcRequest &request, ipc::IpcResponse &response, bool performanceInfo) {
+        auto dataIn{request.inputBuf.at(0)};
+        auto dataOut{request.outputBuf.at(0).cast<opus_int16>()};
+
+        if (dataIn.size() <= sizeof(OpusDataHeader))
+            return result::BadOpusDataHeader;
```
bylaws, line 53: This should throw an exception and so should the other one making this function return void, alternatively make this function return a result and then have all things that call it throw an exception on Nonzero value

bylaws (changes requested):

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+        opus_int32 opusPacketSize{static_cast<opus_int32>(util::SwapEndianness(dataIn.as<OpusDataHeader>().size))};
```
bylaws, line 55: add an accessor to the OpusDataHeader struct to read size in be

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoderManager.h
```diff
@@ -0,0 +1,61 @@
...
+         * @brief Returns the required size for the multi stream decoder's work buffer.
+         * @url https://switchbrew.org/wiki/Audio_services#GetWorkBufferSizeForMultiStream
+         */
+        Result GetWorkBufferSizeForMultiStream(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response);
+
+        SERVICE_DECL(
+            SFUNC(0x0, IHardwareOpusDecoderManager, OpenHardwareOpusDecoder),
+            SFUNC(0x1, IHardwareOpusDecoderManager, GetWorkBufferSize),
+            //SFUNC(0x2, IHardwareOpusDecoderManager, OpenHardwareOpusDecoderForMultiStream),
+            //SFUNC(0x3, IHardwareOpusDecoderManager, GetWorkBufferSizeForMultiStream)
```
PixelyIon, line 58: We don't want unimplemented functions being defined since this just clutters things, while it may not apply as much here. It does for a lot of other services, having it done here and not there ends up being really inconsistent.

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
```diff
@@ -0,0 +1,150 @@
...
+          SFUNC(0x0, IHardwareOpusDecoder, DecodeInterleavedOld),
+          //SFUNC(0x1, IHardwareOpusDecoder, SetContext),
+          //SFUNC(0x2, IHardwareOpusDecoder, DecodeInterleavedForMultiStreamOld),
+          //SFUNC(0x3, IHardwareOpusDecoder, SetContextForMultiStream),
+          SFUNC(0x4, IHardwareOpusDecoder, DecodeInterleavedWithPerfOld),
+          //SFUNC(0x5, IHardwareOpusDecoder, DecodeInterleavedForMultiStreamWithPerfOld),
+          SFUNC(0x6, IHardwareOpusDecoder, DecodeInterleavedWithPerfAndResetOld),
+          //SFUNC(0x7, IHardwareOpusDecoder, DecodeInterleavedForMultiStreamWithPerfAndResetOld),
+          SFUNC(0x8, IHardwareOpusDecoder, DecodeInterleaved),
+          //SFUNC(0x9, IHardwareOpusDecoder, DecodeInterleavedForMultiStream)
```
PixelyIon, line 141: Same as other comment, we don't want unimplemented service functions being defined.

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
```diff
@@ -0,0 +1,150 @@
...
+#include <services/serviceman.h>
+
+namespace skyline::service::codec {
+    /**
+     * @brief This structure is located at the beginning of OpusDataIn for DecodeInterleaved*, with the actual opus packet following this
+     * @note These fields are big-endian
+     * @url https://github.com/switchbrew/libnx/blob/c5a9a909a91657a9818a3b7e18c9b91ff0cbb6e3/nx/include/switch/services/hwopus.h#L19
+     */
+    struct OpusDataHeader {
+        u32 size; //!< Size of the packet following this header
```
PixelyIon, line 18: Explicitly denote the endianess of this member, I'd even recommend making it `sizeBe` so that it wouldn't be used by mistake.

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoderManager.h
```diff
@@ -0,0 +1,61 @@
...
+namespace skyline::service::codec {
+    /**
+     * @brief Struct used to initialize a multi stream decoder
+     * @see opus_multistream_decoder_init()
+     */
+    struct MultiStreamParameters {
+        i32 sampleRate;
+        i32 channelCount;
+        i32 streams; //!< Number of streams
+        i32 coupledStreams; //!< Number of stereo streams
```
PixelyIon, line 17: Change `streams` to `streamCount` and `coupledStreams` to `coupledStreamCount` (or `stereoStreamCount`) and remove the comments.

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoderManager.h
```diff
@@ -0,0 +1,61 @@
...
+
+#pragma once
+
+#include <services/serviceman.h>
+
+namespace skyline::service::codec {
+    /**
+     * @brief Struct used to initialize a multi stream decoder
+     * @see opus_multistream_decoder_init()
+     */
```
PixelyIon, line 12: `Initialization parameters for the Opus multi-stream decoder`

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoderManager.h
```diff
@@ -0,0 +1,61 @@
...
+        i32 sampleRate;
+        i32 channelCount;
+        i32 streams; //!< Number of streams
+        i32 coupledStreams; //!< Number of stereo streams
+        std::array<u8, 0x100> mappings; //!< Array of channel mappings
+    };
+    static_assert(sizeof(MultiStreamParameters) == 0x110);
+
+    /**
+     * @brief This service is used to get an instance of IHardwareOpusDecoder
```
PixelyIon, line 23: We generally discourage the use of `This X` for comments, don't comment on purely usage but more as to what's delegated to the class. I'd say something like `Manages all instances of IHardwareOpusDecoder`/`Manages creation of IHardwareOpusDecoder instances` would be more fitting.

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
```diff
@@ -0,0 +1,150 @@
...
+     * @url https://switchbrew.org/wiki/Audio_services#IHardwareOpusDecoder
+     */
+    class IHardwareOpusDecoder : public BaseService {
+      private:
+        OpusDecoder *decoderState;
+        i32 sampleRate;
+        i32 channelCount;
+        i32 decoderOutputBufferSize;
+
+        void ResetContext();
```
PixelyIon, line 45: A comment would be appreciated denoting that this is resetting the context of the Opus decoder.

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
```diff
@@ -0,0 +1,150 @@
...
+        u32 size; //!< Size of the packet following this header
+        u32 finalRange; //!< Indicates the final range of the codec encoder's entropy coder, this can be left at zero
+
+        u32 GetPacketSize() {
+            return util::SwapEndianness(size);
+        }
+    };
+    static_assert(sizeof(OpusDataHeader) == 0x8);
+
+    size_t CalculateOutBufferSize(i32 sampleRate, i32 channelCount, i32 frameSize);
```
PixelyIon, line 27: Why is this not inside the class (and why doesn't it have a comment)? If you need to use it in the constructor then make it a `static` member (preferably `private` as well).

PixelyIon (changes requested):
Just a few nits here and there, nothing related to functionality.

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
sspacelynx: This is not inside the class and is not static because it is also used in `IHardwareOpusDecoderManager.cpp`, bylaws also asked me about this in some CR above. I'm open to any suggestions if there's a better way to do this, I chose to put that in the `codec` namespace.

sspacelynx: I will add a comment though, missed it.

app/src/main/cpp/skyline/services/timesrv/common.h
```diff
@@ -35,6 +35,10 @@ namespace skyline::service::timesrv {
             return ns;
         }
 
+        constexpr i64 Microseconds() const {
+            return ns / static_cast<i64>(skyline::constant::NsInMicrosecond);
+        }
+
         constexpr i64 Seconds() const {
             return ns / static_cast<i64>(skyline::constant::NsInSecond);
```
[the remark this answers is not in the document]
sspacelynx: I'm not sure what you mean with that, those constants are used elsewhere and simply removing the `skyline` namespace fails to compile.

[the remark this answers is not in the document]
PixelyIon: What I mean is that it's in the `skyline::service::timesrv` namespace so there's no point of scoping into the `skyline` namespace so just `constant::NsInMicrosecond` and so on should be enough.

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
```diff
@@ -0,0 +1,110 @@
...
+     */
+    size_t CalculateOutBufferSize(i32 sampleRate, i32 channelCount, i32 frameSize);
+
+    static constexpr u32 OpusFullbandSampleRate{48000};
+    static constexpr u32 MaxFrameSizeNormal = OpusFullbandSampleRate * 0.040; //!< 40ms frame size limit for normal decoders
+    static constexpr u32 MaxFrameSizeMultiEx = OpusFullbandSampleRate * 0.120; //!< 120ms frame size limit for ex decoders added in 12.0.0
+    static constexpr u32 MaxInputBufferSize{0x600}; //!< Maximum allocated size of the input buffer
+
+    /**
+     * @brief This service decodes an Opus audio stream. Always created with OpenHardwareOpusDecoder by IHardwareOpusDecoderManager
```
PixelyIon, line 39: The information about it's creation doesn't really matter here as it doesn't help you understand what *this* does, after you remove some of the extra fluff you effectively get `Decodes Opus-encoded audio streams` from there you can determine if this comment is even necessary or too obvious. 

A note might help someone who isn't aware about the HW so `@note The Switch has a HW Opus Decoder which this service would interface with, we emulate it using libopus with CPU-decoding`.

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
```diff
@@ -0,0 +1,110 @@
...
+    class IHardwareOpusDecoder : public BaseService {
+      private:
+        std::shared_ptr<kernel::type::KTransferMemory> workBuffer;
+        OpusDecoder *decoderState{};
+        i32 sampleRate;
+        i32 channelCount;
+        i32 decoderOutputBufferSize;
+
+        /**
+         * @brief Resets the codec state to be equivalent to a freshly initialized state
```
PixelyIon, line 51: Just `Resets the Opus decoder's internal state` is enough, no need to specify equivalency to `freshly initialized state`

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
```diff
@@ -0,0 +1,110 @@
...
+         */
+        void ResetContext();
+
+        /**
+         * @brief Called by all DecodeInterleaved* functions
```
PixelyIon, line 56: You made the mistake with describing how the function is used rather than what it does again here, if I want to know more about `DecodeInterleavedImpl` and read the comment which tells me `Called by ...` it's not really useful API-wise.

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
```diff
@@ -0,0 +1,110 @@
...
+         * @param performanceInfo Whether to return decode time taken or not
```
PixelyIon, line 57: `decode time taken` isn't really valid English, `Whether to write the time taken to decode or not` would be more fitting but you could eliminate the comment entirely with a better variable name like `writeDecodeTime` since it's more explicit about what's being done while being the same length

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
```diff
@@ -0,0 +1,110 @@
...
+         */
+        Result DecodeInterleavedImpl(ipc::IpcRequest &request, ipc::IpcResponse &response, bool performanceInfo = false);
+
+      public:
+        /**
+         * @param sampleRate The sample rate of the Opus audio data
+         * @param channelCount The channel count of the Opus audio data
```
PixelyIon, line 64: No need for these comments, they're sort of redundant as it's understood that these values pertain to the stream

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
```diff
@@ -0,0 +1,110 @@
...
+         * @url https://switchbrew.org/wiki/Audio_services#DecodeInterleavedWithPerfOld
+         */
+        Result DecodeInterleavedWithPerfOld(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response);
+
+        /**
+         * @brief Same as DecodeInterleaved
+         * @see DecodeInterleaved
+         * @url https://switchbrew.org/wiki/Audio_services#DecodeInterleaved
+         */
+        Result DecodeInterleavedWithPerfAndResetOld(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response);
```
PixelyIon, line 87: Repetition of the information regarding it being the same is extremely redundant, we won't ever generate Doxygen documentation for Skyline so it's really pointless to have stuff like `@see`. Additionally, don't link to Switchbrew when there isn't a link that directly corresponds to that specific function.

The information covered here in itself is just redundant if it is `Same as DecodeInterleaved` because the point of it would be to cover the differences but if there aren't any then the function in itself becomes redundant and you could just eliminate it and use `SFUNC(0x6, IHardwareOpusDecoder, DecodeInterleaved)` with a comment there `// DecodeInterleavedWithPerfAndResetOld is effectively the same as DecodeInterleaved`.

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
```diff
@@ -0,0 +1,110 @@
...
+
+        /**
+         * @brief Takes an input boolean flag, an OpusDataIn input buffer and a PcmDataOut output buffer.
+         * Decodes the Opus source data to the output buffer and returns decoded data size, decoded sample count and decode time taken in microseconds
+         * @note The bool flag indicates whether or not a reset of the decoder context is being requested
+         * @url https://switchbrew.org/wiki/Audio_services#DecodeInterleaved
+         */
+        Result DecodeInterleaved(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response);
```
PixelyIon, line 95: I'd say make all of these descriptions more oriented towards function rather than purely input/output. The second line actually works fine it's just that the first line seems too much like a schema of inputs so really just chop off the first line for all of them.

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
```diff
@@ -0,0 +1,110 @@
...
+          SFUNC(0x6, IHardwareOpusDecoder, DecodeInterleavedWithPerfAndResetOld),
+          SFUNC(0x8, IHardwareOpusDecoder, DecodeInterleaved),
+      )
+    };
+
+  class OpusException : public skyline::exception {
+    public:
+      OpusException(const std::string& message) : skyline::exception(message) {}
+      OpusException(int errorCode) : skyline::exception("Opus failed with error code {}: {}", errorCode, opus_strerror(errorCode)) {}
+  };
```
PixelyIon, line 109: A comment would be nice `@brief A wrapper around libopus error codes for exceptions` and remove the `std::string` constructor since this should only ever be used with libopus error codes

PixelyIon (changes requested):
You've done a really good job with code but comments are an area that can be improved so I've taken my time to do this review just going through every comment with what would be better. I really don't have any complaints aside from semantics here so I should reiterate that you've done a great job with the rest.

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.cpp
```diff
@@ -0,0 +1,82 @@
...
+        auto dataIn{request.inputBuf.at(0)};
+        auto dataOut{request.outputBuf.at(0).cast<opus_int16>()};
+
+        if (dataIn.size() <= sizeof(OpusDataHeader))
+            throw skyline::exception("Incorrect Opus data size: 0x{:X} (Should be > 0x{:X})", dataIn.size(), sizeof(OpusDataHeader));
+
+        u32 opusPacketSize{dataIn.as<OpusDataHeader>().GetPacketSize()};
+        u32 requiredInSize{static_cast<u32>(opusPacketSize + sizeof(OpusDataHeader))};
+        if (opusPacketSize > MaxInputBufferSize || dataIn.size() < requiredInSize)
+            throw OpusException(OPUS_BUFFER_TOO_SMALL);
```
bylaws, line 62: don't use an opus exception here

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
```diff
@@ -0,0 +1,97 @@
...
+
+namespace skyline::service::codec {
+    /**
+     * @return The required output buffer size for decoding an Opus stream with the given parameters
+     */
+    size_t CalculateOutBufferSize(i32 sampleRate, i32 channelCount, i32 frameSize);
+
+    static constexpr u32 OpusFullbandSampleRate{48000};
+    static constexpr u32 MaxFrameSizeNormal = OpusFullbandSampleRate * 0.040; //!< 40ms frame size limit for normal decoders
+    static constexpr u32 MaxFrameSizeMultiEx = OpusFullbandSampleRate * 0.120; //!< 120ms frame size limit for ex decoders added in 12.0.0
```
bylaws, line 20:
```suggestion
    static constexpr u32 MaxFrameSizeEx = OpusFullbandSampleRate * 0.120; //!< 120ms frame size limit for ex decoders added in 12.0.0
```

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoder.h
```diff
@@ -0,0 +1,97 @@
...
+            SFUNC(0x4, IHardwareOpusDecoder, DecodeInterleavedWithPerfOld),
+            SFUNC(0x6, IHardwareOpusDecoder, DecodeInterleaved), // DecodeInterleavedWithPerfAndResetOld is effectively the same as DecodeInterleaved
+            SFUNC(0x8, IHardwareOpusDecoder, DecodeInterleaved),
+        )
+    };
+
+    /**
+     * @brief A wrapper around libopus error codes for exceptions
+     */
+    class OpusException : public skyline::exception {
```
bylaws, line 93: drop the skyline::

app/src/main/cpp/skyline/services/codec/results.h
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright © 2021 Skyline Team and Contributors (https://github.com/skyline-emu/)
+
+#pragma once
+
+#include <common.h>
+
+namespace skyline::service::codec::result {
+    constexpr Result OpusBadArgument(111, 2);
```
bylaws, line 9: drop this whole file

app/src/main/cpp/skyline/services/codec/IHardwareOpusDecoderManager.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+namespace skyline::service::codec {
+    static size_t CalculateBufferSize(i32 sampleRate, i32 channelCount) {
+        i32 requiredSize{opus_decoder_get_size(channelCount)};
+        requiredSize += MaxInputBufferSize + CalculateOutBufferSize(sampleRate, channelCount, MaxFrameSizeNormal);
+        return requiredSize;
+    }
+
+    Result IHardwareOpusDecoderManager::OpenHardwareOpusDecoder(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
+        auto sampleRate{request.Pop<i32>()};
```
bylaws, line 17: don't use auto here, name the integer types instead

PixelyIon (approved):
Great job with this! It took a few review iterations to get this just right but it's really nice to see that it has reached this point :tada:

bylaws:
Desi girl linx-chan nweeds hwer pweearrr merguwud

.gitignore
```diff
@@ -49,6 +49,7 @@ captures/
 .idea/caches
 .idea/modules.xml
 .idea/navEditor.xml
+.idea/runConfigurations.xml
```
bylaws, line 52: approve

bylaws (approved):

PR #199, opened 2021-07-06
Implement Vulkan Presentation Engine + Rework GraphicBufferProducer & VI

### LINKED ISSUE #197: Vulkan Presentation Engine
We will utilize [Vulkan](https://www.vulkan.org/) as our only GPU backend due to the high feature requirement for Switch GPU emulation. As a result of this our presentation engine must be written to use Vulkan rather than using [ANativeWindow](https://developer.android.com/ndk/reference/group/a-native-window) directly.

### LINKED ISSUE #198: Accurate GraphicBufferProducer Implementation
Our current [GraphicBufferProducer](https://cs.android.com/android/platform/superproject/+/android-5.1.1_r38:frameworks/native/include/gui/IGraphicBufferProducer.h) suffers from several issues due to inaccuracies in it, one of these is completely ignoring any arguments relating to presentation properties such as transformations which can lead to images being rendered in the incorrect orientation, swap interval which can lead to the game being sped-up and certain API-emulation issues such as returning the incorrect errors could lead to crashes.

PixelyIon:
The following items were performed in this PR:
* Implement [Vulkan](https://www.vulkan.org/) [`Instance`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkInstance.html) + [`PhysicalDevice`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPhysicalDevice.html) + [`Device`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkDevice.html) initialization with [`VK_EXT_debug_report`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_EXT_debug_report.html) support
* Integrate [`VK_LAYER_KHRONOS_validation`](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/docs/khronos_validation_layer.md) from Android NDK into debug builds
* Implement base Vulkan infrastructure (`FenceCycle`, Memory Manager [using [VMA](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator)] and Command Scheduler)
* Rework `GuestTexture` + `Texture` classes with Vulkan infrastructure
* Implement Vulkan-based Presentation Engine
* Isolate guest and host presentation for allowing [n-Buffering](https://en.wikipedia.org/wiki/Multiple_buffering) on host
* Utilize [`AChoreographer`](https://developer.android.com/ndk/reference/group/choreographer) for accurate V-Sync events and adaptive refresh rate updates
* Dynamic swap-interval management using [`VK_GOOGLE_display_timing`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_GOOGLE_display_timing.html)
* Implement cropping, transformation and scaling using host [Surfaceflinger](https://source.android.com/devices/graphics/surfaceflinger-windowmanager) for [HWC](https://source.android.com/devices/graphics/hwc) acceleration
* Rework Performance Statistics to provide a more accurate representation of performance
* Introduce `Display` settings for fine-grained control over presentation
* Rework `GraphicBufferProducer` for accuracy and (re)implement the following transactions:
* * `RequestBuffer`
* * `SetBufferCount`
* * `DequeueBuffer`
* * `DetachBuffer`
* * `DetachNextBuffer`
* * `AttachBuffer`
* * `QueueBuffer`
* * `CancelBuffer`
* * `Query`
* * `Connect`
* * `Disconnect`
* * `SetPreallocatedBuffer`
* Rework Host1X Syncpoint for code quality and performance
* Fix race condition in Host1X Syncpoints affecting titles like [Sonic Mania](https://www.nintendo.com/games/detail/sonic-mania-switch/)
* Rework [Display Services (`VI`)](https://switchbrew.org/wiki/Display_services) + [`IHOSBinderDriver`](https://switchbrew.org/wiki/Nvnflinger_services#dispdrv) for accuracy
* Implement SVC [`UnmapSharedMemory`](https://switchbrew.org/wiki/SVC#UnmapSharedMemory)

Closes #197 
Closes #198


app/src/main/cpp/skyline/gpu/texture/texture.h
```diff
@@ -0,0 +1,352 @@
...
+        Texture(GPU &gpu, BackingType &&backing, texture::Dimensions dimensions, const texture::Format &format, vk::ImageLayout layout, vk::ImageTiling tiling, vk::ComponentMapping mapping = {});
+
+        /**
+         * @brief Creates and allocates memory for the backing to creates a texture object wrapping it
+         * @param usage Usage flags that will applied aside from VK_IMAGE_USAGE_TRANSFER_SRC_BIT/VK_IMAGE_USAGE_TRANSFER_DST_BIT which are mandatory
+         */
+        Texture(GPU &gpu, texture::Dimensions dimensions, const texture::Format &format, vk::ImageLayout initialLayout = vk::ImageLayout::eGeneral, vk::ImageUsageFlags usage = {}, vk::ImageTiling tiling = vk::ImageTiling::eOptimal, vk::ComponentMapping mapping = {});
+
+        /**
+         * @brief Acquires an exclusive lock on the texture for the calling thread
```
bylaws, line 266: Note that these are lower case so they can be used with unique_lock

app/src/main/cpp/skyline/kernel/memory.cpp
```diff
@@ -62,7 +62,7 @@ namespace skyline::kernel {
             throw exception("Cannot find a suitable carveout for the guest address space");
 
         auto result{mmap(reinterpret_cast<void *>(base.address), base.size, PROT_NONE, MAP_FIXED | MAP_ANONYMOUS | MAP_PRIVATE, -1, 0)};
-        if (result == MAP_FAILED) [[unlikely]]
+        if (result == MAP_FAILED)
```
bylaws, line 65: I don't understand why you removed this?

app/src/main/cpp/skyline/kernel/types/KSharedMemory.cpp
```diff
@@ -43,15 +43,33 @@ namespace skyline::kernel::type {
...
 
+    void KSharedMemory::Unmap(u8 *ptr, u64 size) {
+        if (!state.process->memory.base.IsInside(ptr) || !state.process->memory.base.IsInside(ptr + size))
+            throw exception("KPrivateMemory allocation isn't inside guest address space: 0x{:X} - 0x{:X}", ptr, ptr + size);
+        if (!util::PageAligned(ptr) || !util::PageAligned(size))
+            throw exception("KSharedMemory mapping isn't page-aligned: 0x{:X} - 0x{:X} (0x{:X})", ptr, ptr + size, size);
+        if (guest.ptr != ptr && guest.size != size)
+            throw exception("Unmapping KSharedMemory partially is not supported: Requested Unmap: 0x{:X} - 0x{:X} (0x{:X}), Current Mapping: 0x{:X} - 0x{:X} (0x{:X})", ptr, ptr + size, size, guest.ptr, guest.ptr + guest.size, guest.size);
+
+        if (mmap(ptr, size, PROT_NONE, MAP_ANONYMOUS | MAP_PRIVATE | MAP_FIXED, 0, 0) == MAP_FAILED)
```
bylaws, line 54: no unlikely here? I don't understand our policy on this

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.cpp
```diff
@@ -1,214 +1,702 @@
...
 #include <services/nvdrv/driver.h>
 #include <services/nvdrv/devices/nvmap.h>
 #include <services/common/fence.h>
 #include "GraphicBufferProducer.h"
 
 namespace skyline::service::hosbinder {
-    Buffer::Buffer(const GbpBuffer &gbpBuffer, std::shared_ptr<gpu::Texture> texture) : gbpBuffer(gbpBuffer), texture(std::move(texture)) {}
+    GraphicBufferProducer::GraphicBufferProducer(const DeviceState &state) : state(state), bufferEvent(std::make_shared<kernel::type::KEvent>(state, true)) {}
+
+    u8 GraphicBufferProducer::GetPendingBufferCount() {
```
bylaws, line 17: Any reason for a `u8` here when they're more expensive

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.cpp
```diff
@@ -1,214 +1,702 @@
...
+        return AndroidStatus::Ok;
     }
 
-    void GraphicBufferProducer::CancelBuffer(Parcel &in) {
-        u32 slot{in.Pop<u32>()};
-        //auto fences{in.Pop<std::array<nvdrv::Fence, 4>>()};
+    AndroidStatus GraphicBufferProducer::DetachNextBuffer(std::optional<GraphicBuffer> &graphicBuffer, std::optional<AndroidFence> &fence) {
+        std::scoped_lock lock(mutex);
+        auto bufferSlot{queue.end()};
+        for (auto it{queue.begin()}; it != queue.end(); it++) {
```
bylaws, line 174: Could find_if work here?

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.cpp
```diff
@@ -1,214 +1,702 @@
...
+        }
+
+        if (graphicBuffer.magic != GraphicBuffer::Magic)
+            throw exception("Unexpected GraphicBuffer magic: 0x{} (Expected: 0x{})", graphicBuffer.magic, GraphicBuffer::Magic);
+        else if (graphicBuffer.intCount != sizeof(NvGraphicHandle) / sizeof(u32))
+            throw exception("Unexpected GraphicBuffer native_handle integer count: 0x{} (Expected: 0x{})", graphicBuffer.intCount, sizeof(NvGraphicHandle));
+
+        auto &handle{graphicBuffer.graphicHandle};
+        if (handle.magic != NvGraphicHandle::Magic)
+            throw exception("Unexpected NvGraphicHandle magic: {}", handle.surfaceCount);
```
bylaws, line 217: I too have a surfacecount magic

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.cpp
```diff
@@ -1,214 +1,702 @@
...
+        else if (handle.surfaceCount < 1)
+            throw exception("At least one surface is required in a buffer: {}", handle.surfaceCount);
+        else if (handle.surfaceCount > 1)
+            throw exception("Multi-planar surfaces are not supported: {}", handle.surfaceCount);
+
+        auto &surface{graphicBuffer.graphicHandle.surfaces.at(0)};
+        if (surface.scanFormat != NvDisplayScanFormat::Progressive)
```
bylaws, line 224: 540i interlaced CRT skyline :(

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.cpp
```diff
@@ -1,214 +1,702 @@
...
+
+        bufferSlot->state = BufferState::Dequeued;
+        bufferSlot->wasBufferRequested = true;
+        bufferSlot->isPreallocated = false;
+        bufferSlot->graphicBuffer = std::make_unique<GraphicBuffer>(graphicBuffer);
+
+        slot = std::distance(queue.begin(), bufferSlot);
+
+        preallocatedBufferCount = std::count_if(queue.begin(), queue.end(), [](const BufferSlot &slot) { return slot.graphicBuffer && slot.isPreallocated; });
+        activeSlotCount = std::count_if(queue.begin(), queue.end(), [](const BufferSlot &slot) { return static_cast<bool>(slot.graphicBuffer); });
```
bylaws, line 237:
```suggestion
        activeSlotCount = std::count_if(queue.begin(), queue.end(), [](const BufferSlot &slot) { return slot.graphicBuffer != nullptr; });
```

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.cpp
```diff
@@ -1,214 +1,702 @@
...
+
+        state.logger->Debug("#{} - Dimensions: {}x{} [Stride: {}], Format: {}, Layout: {}, {}: {}, Usage: 0x{:X}, NvMap {}: {}, Buffer Start/End: 0x{:X} -> 0x{:X}", slot, surface.width, surface.height, handle.stride, ToString(graphicBuffer.format), ToString(surface.layout), surface.layout == NvSurfaceLayout::Blocklinear ? "Block Height" : "Pitch", surface.layout == NvSurfaceLayout::Blocklinear ? 1U << surface.blockHeightLog2 : surface.pitch, graphicBuffer.usage, surface.nvmapHandle ? "Handle" : "ID", surface.nvmapHandle ? surface.nvmapHandle : handle.nvmapId, surface.offset, surface.offset + surface.size);
```
bylaws, line 239: wrap this line

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.cpp
```diff
@@ -1,214 +1,702 @@
...
+            if (surface.size > (nvBuffer->size - surface.offset))
+                throw exception("Surface doesn't fit into NvMap mapping of size 0x{:X} when mapped at 0x{:X} -> 0x{:X}", nvBuffer->size, surface.offset, surface.offset + surface.size);
+
+            gpu::texture::TileMode tileMode;
+            gpu::texture::TileConfig tileConfig{};
+            if (surface.layout == NvSurfaceLayout::Blocklinear) {
+                tileMode = gpu::texture::TileMode::Block;
+                tileConfig = {
+                    .surfaceWidth = static_cast<u16>(surface.width),
+                    .blockHeight = static_cast<u8>(1U << surface.blockHeightLog2),
```
bylaws, line 335: blockHeightBits is a much better varname for surface.blockHeightLog2

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.cpp
```diff
@@ -1,214 +1,702 @@
...
+        buffer.wasBufferRequested = false;
+        buffer.isPreallocated = static_cast<bool>(graphicBuffer);
+        buffer.graphicBuffer = graphicBuffer ? std::make_unique<GraphicBuffer>(*graphicBuffer) : nullptr;
+        buffer.texture = {};
+
+        if (graphicBuffer) {
+            if (graphicBuffer->magic != GraphicBuffer::Magic)
+                throw exception("Unexpected GraphicBuffer magic: 0x{} (Expected: 0x{})", graphicBuffer->magic, GraphicBuffer::Magic);
+            else if (graphicBuffer->intCount != sizeof(NvGraphicHandle) / sizeof(u32))
+                throw exception("Unexpected GraphicBuffer native_handle integer count: 0x{} (Expected: 0x{})", graphicBuffer->intCount, sizeof(NvGraphicHandle));
```
bylaws, line 558:
```suggestion
                throw exception("Unexpected GraphicBuffer native_handle integer count: 0x{} (Expected: 0x{})", graphicBuffer->intCount, sizeof(NvGraphicHandle) / sizeof(u32));
```

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.cpp
```diff
@@ -1,214 +1,702 @@
...
+            else if (surface.layout == NvSurfaceLayout::Tiled)
+                throw exception("Legacy 16Bx16 tiled surfaces are not supported");
+
+            state.logger->Debug("#{} - Dimensions: {}x{} [Stride: {}], Format: {}, Layout: {}, {}: {}, Usage: 0x{:X}, NvMap {}: {}, Buffer Start/End: 0x{:X} -> 0x{:X}", slot, surface.width, surface.height, handle.stride, ToString(graphicBuffer->format), ToString(surface.layout), surface.layout == NvSurfaceLayout::Blocklinear ? "Block Height" : "Pitch", surface.layout == NvSurfaceLayout::Blocklinear ? 1U << surface.blockHeightLog2 : surface.pitch, graphicBuffer->usage, surface.nvmapHandle ? "Handle" : "ID", surface.nvmapHandle ? surface.nvmapHandle : handle.nvmapId, surface.offset, surface.offset + surface.size);
+        } else {
+            state.logger->Debug("#{} - No GraphicBuffer", slot);
+        }
+
+        preallocatedBufferCount = std::count_if(queue.begin(), queue.end(), [](const BufferSlot &slot) { return slot.graphicBuffer && slot.isPreallocated; });
+        activeSlotCount = std::count_if(queue.begin(), queue.end(), [](const BufferSlot &slot) { return static_cast<bool>(slot.graphicBuffer); });
```
bylaws, line 580:
```suggestion
        activeSlotCount = std::count_if(queue.begin(), queue.end(), [](const BufferSlot &slot) { return slot.graphicBuffer != nullptr; });
```

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.cpp
```diff
@@ -1,214 +1,702 @@
...
+            case TransactionCode::Query: {
+                u32 queryOut{};
+                auto result{Query(in.Pop<NativeWindowQuery>(), queryOut)};
+                out.Push(queryOut);
+                out.Push(result);
+                break;
+            }
+
+            case TransactionCode::Connect: {
+                auto hasProducerListener{static_cast<bool>(in.Pop<u32>())};
```
bylaws, line 671:
```suggestion
                bool hasProducerListener{in.Pop<u32>() != 0};
```

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.h
```diff
@@ -1,160 +1,191 @@
...
+     * @url https://cs.android.com/android/platform/superproject/+/android-5.1.1_r38:frameworks/native/libs/gui/BufferQueueCore.cpp
      */
     class GraphicBufferProducer {
       private:
         const DeviceState &state;
-        std::unordered_map<u32, std::shared_ptr<Buffer>> queue; //!< A vector of shared pointers to all the queued buffers
+        std::mutex mutex; //!< Synchronizes access to the buffer queue
+        constexpr static u8 MaxSlotCount{16}; //!< The maximum amount of buffer slots that a buffer queue can hold, Android supports 64 but they go unused for applications like games so we've lowered this to 16 (https://cs.android.com/android/platform/superproject/+/android-5.1.1_r38:frameworks/native/include/gui/BufferQueueDefs.h;l=29)
+        std::array<BufferSlot, MaxSlotCount> queue;
+        u8 activeSlotCount{}; //!< The amount of slots in the queue that can be dequeued
```
bylaws, line 62: use an i32 for these things, android does so it makes sense to

app/src/main/cpp/skyline/services/hosbinder/IHOSBinderDriver.cpp
```diff
@@ -7,43 +7,166 @@
...
-        request.Skip<u32>();
-        auto addVal{request.Pop<i32>()};
-        auto type{request.Pop<i32>()};
-        state.logger->Debug("Reference Change: {} {} reference", addVal, type ? "strong" : "weak");
+        auto binderHandle{request.Pop<u32>()};
+        if (binderHandle != DefaultBinderLayerHandle)
+            throw exception("Adjusting Binder object reference count for unknown object: #{}", binderHandle);
+
+        auto value{request.Pop<i32>()};
+        bool isStrong{static_cast<bool>(request.Pop<u32>())};
```
bylaws, line 40:
```suggestion
        bool isStrong{request.Pop<u32>() != 0};
```

app/src/main/cpp/skyline/services/hosbinder/IHOSBinderDriver.h
```diff
@@ -3,17 +3,46 @@
...
+    })
 
     /**
      * @brief nvnflinger:dispdrv or nns::hosbinder::IHOSBinderDriver is a translation layer between Android Binder IPC and HOS IPC to communicate with the Android display stack
      */
     class IHOSBinderDriver : public BaseService {
       private:
-        std::shared_ptr<GraphicBufferProducer> producer;
+        DisplayId displayId{DisplayId::Null}; //!< The ID of the display that the layer is connected to
+        constexpr static u64 DefaultLayerId{1}; //!< The VI ID of the default (and only) layer in our surface stack
```
bylaws, line 36: can you not mix the constants like this, add some spacing or smth

app/src/main/cpp/skyline/services/hosbinder/parcel.h
```diff
@@ -67,12 +103,9 @@ namespace skyline::service {
...
         void PushObject(const ObjectType &object) {
-            objects.reserve(objects.size() + sizeof(ObjectType));
-            auto item{reinterpret_cast<const u8 *>(&object)};
-            for (size_t index{}; sizeof(ObjectType) > index; index++) {
-                objects.push_back(*object);
-                item++;
-            }
+            auto offset{objects.size()};
+            objects.resize(offset + sizeof(ObjectType));
+            *(reinterpret_cast<ObjectType *>(objects.data() + offset)) = object;
```
bylaws, line 108: I think you may as well use memcpy here? like it's a bit more understandable

app/src/main/java/emu/skyline/MainActivity.kt
```diff
@@ -120,15 +122,25 @@ class MainActivity : AppCompatActivity() {
...
+            post { setDistanceToTriggerSync(binding.swipeRefreshLayout.height / 3) }
             setOnRefreshListener { loadRoms(false) }
         }
 
         viewModel.stateData.observe(this, ::handleState)
         loadRoms(!settings.refreshRequired)
 
         binding.searchBar.apply {
-            binding.logIcon.setOnClickListener { startActivity(Intent(context, LogActivity::class.java)) }
+            binding.logIcon.setOnClickListener {
```
bylaws, line 133: I think this would look better if it were written in ☕ 🧻

bylaws (changes requested):

app/build.gradle
```diff
@@ -15,7 +15,7 @@ android {
         targetSdkVersion 30
 
         versionCode 3
-        versionName "0.3"
+        versionName "0.0.3"
```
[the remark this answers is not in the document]
PixelyIon: I did this since we need a 3 component version for Vulkan, I did make it `0.0.3` rather than `0.3.0` by mistake but does this matter? It's sort of abstract anyway but I could change it to that if you really want it to be `0.3.0`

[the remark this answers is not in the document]
bylaws: personally prefer 1.0.0 since that's pretty neat

app/src/main/cpp/skyline/gpu/command_scheduler.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+        auto result{(*gpu.vkDevice).allocateCommandBuffers(&commandBufferAllocateInfo, &commandBuffer, *gpu.vkDevice.getDispatcher())};
+        if (result != vk::Result::eSuccess)
+            vk::throwResultException(result, __builtin_FUNCTION());
+        return ActiveCommandBuffer(commandBuffers.emplace_back(gpu.vkDevice, commandBuffer, vkCommandPool));
+    }
+
+    void CommandScheduler::SubmitCommandBuffer(const vk::raii::CommandBuffer &commandBuffer, vk::Fence fence) {
+        std::lock_guard lock(gpu.queueMutex);
+        gpu.vkQueue.submit(vk::SubmitInfo{
```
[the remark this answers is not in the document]
PixelyIon: What would I check? It doesn't return anything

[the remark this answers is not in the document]
bylaws: I didn't know that

app/src/main/cpp/skyline/gpu/command_scheduler.cpp
```diff
@@ -0,0 +1,53 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright © 2021 Skyline Team and Contributors (https://github.com/skyline-emu/)
+
+#include <gpu.h>
+#include "command_scheduler.h"
+
+namespace skyline::gpu {
+    CommandScheduler::CommandBufferSlot::CommandBufferSlot(vk::raii::Device &device, vk::CommandBuffer commandBuffer, vk::raii::CommandPool &pool) : active(true), device(device), commandBuffer(device, commandBuffer, pool), fence(device, vk::FenceCreateInfo{}), cycle(std::make_shared<FenceCycle>(device, *fence)) {}
+
+    bool CommandScheduler::CommandBufferSlot::AllocateIfFree(CommandScheduler::CommandBufferSlot &slot) {
```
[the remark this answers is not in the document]
PixelyIon: `CommandBufferSlot` is `private` so this isn't possible, I don't want to make it public either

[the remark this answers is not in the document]
bylaws: Oh I see

app/src/main/cpp/skyline/gpu/fence_cycle.h
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        ~FenceCycle() {
+            Wait();
+        }
+
+        /**
+         * @brief Wait on a fence cycle till it has been signalled
+         */
+        void Wait() {
```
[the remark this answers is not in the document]
PixelyIon: They're all tiny functions, we generally end up keeping these in headers, the entire class is a pretty minimal wrapper over a Vulkan fence so it doesn't introduce much in the way of compilation overhead (Esp. considering that Vulkan-Hpp is also included)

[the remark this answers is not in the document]
bylaws: yes but literally everything vulkan includes this so move them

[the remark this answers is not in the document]
PixelyIon: It's a negligible amount of overhead, it's written almost as an extension to Vulkan-Hpp

app/src/main/cpp/skyline/gpu/fence_cycle.h
```diff
@@ -0,0 +1,134 @@
...
+            }
+        }
+
+      public:
+        FenceCycle(const vk::raii::Device &device, vk::Fence fence) : signalled(false), device(device), fence(fence) {
+            device.resetFences(fence);
+        }
+
+        ~FenceCycle() {
+            Wait();
```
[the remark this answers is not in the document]
PixelyIon: If that happens then the GPU driver will likely throw an error that isn't timeout-related, we should depend on that rather than introduce an arbitrary timeout

app/src/main/cpp/skyline/gpu/fence_cycle.h
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        /**
+         * @brief A version of AttachObject optimized for several objects being attached at once
+         */
+        void AttachObjects(std::initializer_list<std::shared_ptr<FenceCycleDependency>> dependencies) {
+            if (!signalled.test(std::memory_order_consume)) {
+                auto it{dependencies.begin()}, next{std::next(it)};
+                if (it != dependencies.end()) {
+                    while (next != dependencies.end()) {
```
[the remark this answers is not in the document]
PixelyIon: I think `for`
```cpp
for (; next != dependencies.end(); next++) {
    (*it)->next = *next;
    it = next;
}
```
looks worse than `while`
```cpp
while (next != dependencies.end()) {
    (*it)->next = *next;
    it = next;
    next = std::next(next);
}
```
but I don't want to bikeshed so I'll let you have it your way

app/src/main/cpp/skyline/gpu/memory_manager.cpp
```diff
@@ -0,0 +1,126 @@
...
+    Image::~Image() {
+        if (vmaAllocator && vmaAllocation && vkImage) {
+            if (pointer)
+                vmaUnmapMemory(vmaAllocator, vmaAllocation);
+            vmaDestroyImage(vmaAllocator, vkImage, vmaAllocation);
+        }
+    }
+
+    u8 *Image::data() {
+        if (pointer) [[likely]]
```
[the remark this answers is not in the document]
PixelyIon: It will only not be the case on the first invocation, all further invocations should have the `Image` allocated on the CPU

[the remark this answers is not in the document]
bylaws: I mean isn't it a forward jump? is it realy needed here

app/src/main/cpp/skyline/gpu/memory_manager.cpp
```diff
@@ -0,0 +1,126 @@
...
+        };
+        VmaAllocatorCreateInfo allocatorCreateInfo{
+            .physicalDevice = *gpu.vkPhysicalDevice,
+            .device = *gpu.vkDevice,
+            .instance = *gpu.vkInstance,
+            .pVulkanFunctions = &vulkanFunctions,
+            .vulkanApiVersion = GPU::VkApiVersion,
+        };
+        ThrowOnFail(vmaCreateAllocator(&allocatorCreateInfo, &vmaAllocator));
+        // TODO: Use VK_KHR_dedicated_allocation when available (Should be on Adreno GPUs)
```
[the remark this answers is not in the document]
PixelyIon: It'll be done later, I need to implement a Vulkan quirks system where it registers optional features/extensions/other device-specific quirks

app/src/main/cpp/skyline/gpu/memory_manager.cpp
```diff
@@ -0,0 +1,126 @@
...
+        };
+        VmaAllocationCreateInfo allocationCreateInfo{
+            .flags = VMA_ALLOCATION_CREATE_MAPPED_BIT,
+            .usage = VMA_MEMORY_USAGE_CPU_ONLY,
+        };
+
+        VkBuffer buffer;
+        VmaAllocation allocation;
+        VmaAllocationInfo allocationInfo;
+        ThrowOnFail(vmaCreateBuffer(vmaAllocator, &static_cast<const VkBufferCreateInfo &>(bufferCreateInfo), &allocationCreateInfo, &buffer, &allocation, &allocationInfo));
```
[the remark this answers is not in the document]
PixelyIon: Sadly, it cannot as the cast from `vk::ImageCreateInfo` (Vulkan-Hpp type) to `VkImageCreateInfo` (Vulkan C type) is done by defining `vk::ImageCreateInfo::operator VkImageCreateInfo &()` which needs to be casted to using an explicit cast prior to taking a pointer value since the pointer value becomes `vk::ImageCreateInfo*` that cannot be implicitly casted to `VkImageCreateInfo*` as pointer values do not inherit the implicit casts of their underlying types.

app/src/main/cpp/skyline/gpu/memory_manager.h
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * @brief A view into a CPU mapping of a Vulkan buffer
+     * @note The mapping **should not** be used after the lifetime of the object has ended
+     */
+    struct StagingBuffer : public span<u8>, public FenceCycleDependency {
+        VmaAllocator vmaAllocator;
+        VmaAllocation vmaAllocation;
+        vk::Buffer vkBuffer;
+
+        constexpr StagingBuffer(u8 *pointer, size_t size, VmaAllocator vmaAllocator, vk::Buffer vkBuffer, VmaAllocation vmaAllocation) : vmaAllocator(vmaAllocator), vkBuffer(vkBuffer), vmaAllocation(vmaAllocation), span(pointer, size) {}
```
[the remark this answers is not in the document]
PixelyIon: I wouldn't

app/src/main/cpp/skyline/gpu/native_window.h
```diff
@@ -0,0 +1,303 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright © 2021 Skyline Team and Contributors (https://github.com/skyline-emu/)
+// Copyright © 2021 The Android Open Source Project
+
+#pragma once
+
```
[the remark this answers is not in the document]
PixelyIon: We're not having file comments, they're inconsistent and weird to have because of how blatantly simple file-based categorization generally ends up being like you could easily collect that information in the time.. I didn't realize this was the native window file I agree now but I took a lot of time to write up this comment so I will post it regardless so you have an equivalent amount of time wasted by reading it.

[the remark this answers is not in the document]
bylaws: 😆

app/src/main/cpp/skyline/gpu/presentation_engine.cpp
```diff
@@ -2,83 +2,288 @@
...
+            if (engine->window)
+                engine->window->perform(engine->window, NATIVE_WINDOW_GET_REFRESH_CYCLE_DURATION, &engine->refreshCycleDuration);
+            else
+                engine->refreshCycleDuration = cycleLength;
+        }
+
+        engine->lastChoreographerTime = frameTimeNanos;
+        engine->vsyncEvent->Signal();
+
+        AChoreographer_postFrameCallback(AChoreographer_getInstance(), reinterpret_cast<AChoreographer_frameCallback>(&ChoreographerCallback), engine);
```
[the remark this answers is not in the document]
PixelyIon: We cannot use it as [`AChoreographer_postFrameCallback64`](https://developer.android.com/ndk/reference/group/choreographer#achoreographer_postframecallback64) is only available on API version 29 or later, I would have to have `dlopen` with branching depending on the API version. I could also up the API version I suppose as we already sort of decided this but I didn't go through with it?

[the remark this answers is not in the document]
bylaws: up the API version then :)

app/src/main/cpp/skyline/gpu/presentation_engine.cpp
```diff
@@ -2,83 +2,288 @@
...
+            });
+            if (!gpu.vkPhysicalDevice.getSurfaceSupportKHR(gpu.vkQueueFamilyIndex, **vkSurface))
+                throw exception("Vulkan Queue doesn't support presentation with surface");
+            vkSurfaceCapabilities = gpu.vkPhysicalDevice.getSurfaceCapabilitiesKHR(**vkSurface);
+
+            if (swapchainExtent && swapchainFormat)
+                UpdateSwapchain(swapchainFormat, swapchainExtent);
+
+            if (window->common.magic != AndroidNativeWindowMagic)
+                throw exception("ANativeWindow* has unexpected magic: {} instead of {}", span(&window->common.magic, 1).as_string(true), span<const u8>(reinterpret_cast<const u8 *>(&AndroidNativeWindowMagic), sizeof(u32)).as_string(true));
```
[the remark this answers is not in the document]
PixelyIon: It's not null-terminated rather *null or `sizeof(u32)`* whichever comes first, `libfmt` doesn't handle something having multiple null terminators in a single string and ends up passing it through to the final string which then does an incomplete print

app/src/main/cpp/skyline/gpu/presentation_engine.cpp
```diff
@@ -2,83 +2,288 @@
...
+            if (window->common.version != sizeof(ANativeWindow))
+                throw exception("ANativeWindow* has unexpected version: {} instead of {}", window->common.version, sizeof(ANativeWindow));
+
+            if (windowCrop)
+                window->perform(window, NATIVE_WINDOW_SET_CROP, &windowCrop);
+            if (windowScalingMode != NativeWindowScalingMode::ScaleToWindow)
+                window->perform(window, NATIVE_WINDOW_SET_SCALING_MODE, static_cast<i32>(windowScalingMode));
+
+            window->perform(window, NATIVE_WINDOW_ENABLE_FRAME_TIMESTAMPS, true);
```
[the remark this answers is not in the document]
PixelyIon: I forgot to add one, good catch!

app/src/main/cpp/skyline/gpu/texture/texture.h
PixelyIon: See the class comment `@note This class conforms to the Lockable and BasicLockable C++ named requirements`

app/src/main/cpp/skyline/kernel/memory.cpp
PixelyIon: It's implicit for any paths that throw an exception, the compiler automatically understands that they're unlikely branches

app/src/main/cpp/skyline/kernel/types/KSharedMemory.cpp
PixelyIon: Same as above, our policy is effectively `[[unlikely]]` on any unlikely path that isn't implicitly marked as such

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.cpp
PixelyIon: See how it doesn't `break` after finding one, it needs to find the oldest `Free` slot rather than just any `Free` slot in the queue

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.cpp
PixelyIon: In my defense, a lot of copy pasting was involved :stuck_out_tongue:

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.cpp
PixelyIon: I'll let you get Khronos Group to support it in Vulkan first then I can add support for interlaced output to Skyline

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.cpp
PixelyIon: The units involved aren't bits though, height isn't measured in bits and I think `log2` is consistent with the TRM

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.h
PixelyIon: I use `unsigned` as a guarantee here, the code will break very hard if any of these values are negative

bylaws: well im sure android's code would too? if not use u32

PixelyIon: There's no need to make the type wider either since `MaxSlotCount` is `u8` and this is guaranteed to never go above that, all this would do is make the loads/stores go from `LDRB`/`STRB` to `LDR`/`STR` which are marginally less efficient but for no reason, none of the arithmetic on the CPU side uses this directly so no instructions will be generated from the usage of it here.

app/src/main/java/emu/skyline/MainActivity.kt
PixelyIon: Two people disagree so you're clearly wrong here, it looks the best in Kotlin :sunglasses:

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.h
bylaws: but android itself supports more slots than a u8 no?

PixelyIon: It actually doesn't, Android just likes to be more inefficient I suppose?

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.cpp
bylaws: So reversed find_if?

.idea/jarRepositories.xml
```diff
@@ -36,5 +36,10 @@
       <option name="name" value="MavenLocal" />
```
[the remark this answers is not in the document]
bylaws: Not doing this?

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.h
bylaws: Word

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.cpp
PixelyIon: What if the first slot is the oldest? The queue isn't ordered by age

app/src/main/cpp/skyline/services/hosbinder/GraphicBufferProducer.h
PixelyIon: > none of the arithmetic on the CPU side uses this directly so no instructions will be generated from the usage of it here.
Anyway, this is sort of pedantic so we'll address it if it comes up in the profiler till then we can take advantage of the lower memory footprint :P

.idea/jarRepositories.xml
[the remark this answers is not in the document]
bylaws: This file still hasn't been deleted

bylaws:
> The following items were performed in this PR:
> * Implement [Vulkan](https://www.vulkan.org/) [`Instance`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkInstance.html) + [`PhysicalDevice`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPhysicalDevice.html) + [`Device`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkDevice.html) initialization with [`VK_EXT_debug_report`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_EXT_debug_report.html) support
> * Integrate [`VK_LAYER_KHRONOS_validation`](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/docs/khronos_validation_layer.md) from Android NDK into debug builds
> * Implement base Vulkan infrastructure (`FenceCycle`, Memory Manager [using [VMA](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator)] and Command Scheduler)
> * Rework `GuestTexture` + `Texture` classes with Vulkan infrastructure
> * Implement Vulkan-based Presentation Engine
> * Isolate guest and host presentation for allowing [n-Buffering](https://en.wikipedia.org/wiki/Multiple_buffering) on host
> * Utilize [`AChoreographer`](https://developer.android.com/ndk/reference/group/choreographer) for accurate V-Sync events and adaptive refresh rate updates
> * Dynamic swap-interval management using [`VK_GOOGLE_display_timing`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_GOOGLE_display_timing.html)
> * Implement cropping, transformation and scaling using host [Surfaceflinger](https://source.android.com/devices/graphics/surfaceflinger-windowmanager) for [HWC](https://source.android.com/devices/graphics/hwc) acceleration
> * Rework Performance Statistics to provide a more accurate representation of performance
> * Introduce `Display` settings for fine-grained control over presentation
> * Rework `GraphicBufferProducer` for accuracy and (re)implement the following transactions:
> * * `RequestBuffer`
> * * `SetBufferCount`
> * * `DequeueBuffer`
> * * `DetachBuffer`
> * * `DetachNextBuffer`
> * * `AttachBuffer`
> * * `QueueBuffer`
> * * `CancelBuffer`
> * * `Query`
> * * `Connect`
> * * `Disconnect`
> * * `SetPreallocatedBuffer`
> * Rework Host1X Syncpoint for code quality and performance
> * Fix race condition in Host1X Syncpoints affecting titles like [Sonic Mania](https://www.nintendo.com/games/detail/sonic-mania-switch/)
> * Rework [Display Services (`VI`)](https://switchbrew.org/wiki/Display_services) + [`IHOSBinderDriver`](https://switchbrew.org/wiki/Nvnflinger_services#dispdrv) for accuracy
> * Implement SVC [`UnmapSharedMemory`](https://switchbrew.org/wiki/SVC#UnmapSharedMemory)
> 
> Closes #197 
> Closes #198

I approve

bylaws (approved):
Assuming you delete jarRepositories

PR #210, opened 2021-07-16
Csrng service basic implementation

sepcnt:


PixelyIon (changes requested):
There's a lot wrong with this PR, I recommend taking a look at our [Contributing Guidelines](https://github.com/skyline-emu/skyline/blob/master/CONTRIBUTING.md) and other service implementations (and just Skyline C++ code in general) since there's a lot of style errors and oddities in this PR. I would generally point these out in a CR but we want to get a rough metric of competency since we can't really guide you through the PR fully as that removes the point of additional contributors.

PR #212, opened 2021-07-17
SPL service basic implementation

sepcnt:
Correct code style and standardized

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -215,6 +215,11 @@ add_library(skyline SHARED
         ${source_DIR}/skyline/services/ssl/ISslContext.cpp
         ${source_DIR}/skyline/services/prepo/IPrepoService.cpp
         ${source_DIR}/skyline/services/mmnv/IRequest.cpp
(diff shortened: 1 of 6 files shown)


PixelyIon (changes requested):
It is better than before (#210) but there's still too much wrong for such a small service, I'm sorry but Skyline is a project where we expect really high code-quality alongside well-written comments.

A lot of it comes from being able to introspect into existing code and apply that to new code, we would love to have additional contributors but I think in this case there's a bit too much that you've gotten wrong which makes it stick out from the rest of the codebase.

I think that Skyline might not be a project that might be the right fit for you to contribute to, this doesn't mean you're a bad programmer by any means, just that we end up requiring a lot of things to be a certain way and can't invest much time in teaching people these since it takes away from development time. 

We're very oriented towards long-term contributors and little oversight wrt design decisions and code-style which I feel won't really be achieved in this case. I do still want to thank you for attempt at contributing and if @bylaws or someone else does want to push this to a state where it can be merged I don't mind doing so but I just wanted to state my position on the matter.

PR #220, opened 2021-08-11
csrng service implementation

lynxnb:
See individual commits description.

### DIFF
--- a/app/src/main/cpp/skyline/common.h
+++ b/app/src/main/cpp/skyline/common.h
@@ -340,6 +340,20 @@ namespace skyline {
             return frz::elsa<frz::string>{}(frz::string(view.data(), view.size()), 0);
         }
 
+        /**
+         * @brief Selects the largest possible integer type for representing an object alongside providing the size of the object in terms of the underlying type
+         */
+        template<class T>
+        struct IntegerFor {
+            using type = std::conditional_t<sizeof(T) % sizeof(u64) == 0, u64,
+                std::conditional_t<sizeof(T) % sizeof(u32) == 0, u32,
+                    std::conditional_t<sizeof(T) % sizeof(u16) == 0, u16, u8>
+                >
+            >;
+
+            static constexpr size_t count{sizeof(T) / sizeof(type)};
+        };
+
         namespace detail {
             static thread_local std::mt19937_64 generator{GetTimeTicks()};
         }
@@ -353,6 +367,11 @@ namespace skyline {
             std::independent_bits_engine<std::mt19937_64, std::numeric_limits<T>::digits, T> gen(detail::generator);
 
             std::generate(in.begin(), in.end(), gen);
+        }
+
+        template<class T> requires (!std::is_integral_v<T> && std::is_trivially_copyable_v<T>)
+        void FillRandomBytes(T &object) {
+            FillRandomBytes(std::span(reinterpret_cast<typename IntegerFor<T>::type *>(&object), IntegerFor<T>::count));
         }
     }
 
--- a/app/src/main/cpp/skyline/common.h
+++ b/app/src/main/cpp/skyline/common.h
@@ -22,6 +22,7 @@
 #include <memory>
 #include <compare>
 #include <variant>
+#include <random>
 #include <sys/mman.h>
 #include <fmt/format.h>
 #include <frozen/unordered_map.h>
@@ -337,6 +338,21 @@ namespace skyline {
          */
         constexpr std::size_t Hash(std::string_view view) {
             return frz::elsa<frz::string>{}(frz::string(view.data(), view.size()), 0);
+        }
+
+        namespace detail {
+            static thread_local std::mt19937_64 generator{GetTimeTicks()};
+        }
+
+        /**
+         * @brief Fills an array with random data from a Mersenne Twister pseudo-random generator
+         * @note The generator is seeded with the the current time in ticks
+         */
+        template<typename T> requires (std::is_integral_v<T>)
+        void FillRandomBytes(std::span<T> in) {
+            std::independent_bits_engine<std::mt19937_64, std::numeric_limits<T>::digits, T> gen(detail::generator);
+
+            std::generate(in.begin(), in.end(), gen);
         }
     }
 
--- a/app/src/main/cpp/skyline/services/spl/IRandomInterface.cpp
+++ b/app/src/main/cpp/skyline/services/spl/IRandomInterface.cpp
@@ -1,0 +1,17 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright © 2021 Skyline Team and Contributors (https://github.com/skyline-emu/)
+
+#include "IRandomInterface.h"
+
+#include <common.h>
+
+namespace skyline::service::spl {
+    IRandomInterface::IRandomInterface(const DeviceState &state, ServiceManager &manager) : BaseService(state, manager) {}
+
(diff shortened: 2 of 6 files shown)


app/src/main/cpp/skyline/services/spl/IRandomInterface.h
```diff
@@ -0,0 +1,26 @@
...
+    /**
+     * @brief csrng provides cryptographically secure random number generation
+     * @url https://switchbrew.org/wiki/SPL_services#csrng
+     */
+    class IRandomInterface : public BaseService {
+      public:
+        IRandomInterface(const DeviceState &state, ServiceManager &manager);
+
+        /**
+         * @brief Fills an input buffer with random bytes
```
bylaws, line 18: Output Buffer

app/src/main/cpp/skyline/common.h
```diff
@@ -338,6 +339,18 @@ namespace skyline {
         constexpr std::size_t Hash(std::string_view view) {
             return frz::elsa<frz::string>{}(frz::string(view.data(), view.size()), 0);
         }
+
+        /**
+         * @brief Fills an array with random data from a Mersenne Twister pseudo-random generator
+         * @note Currently the generator is not seeded so produced values will be identical between two runs
+         */
+        template<typename T>
```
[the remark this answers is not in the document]
sspacelynx: Was just looking into the syntax required to do this

app/src/main/cpp/skyline/services/spl/IRandomInterface.h
sspacelynx: whoops

app/src/main/cpp/skyline/common.h
```diff
@@ -22,6 +22,7 @@
 #include <memory>
 #include <compare>
 #include <variant>
+#include <random>
```
PixelyIon, line 25: You can remove the `#include <random>` from `common/uuid.cpp` which now has it redundantly included now that it's included here.

app/src/main/cpp/skyline/common.h
```diff
@@ -338,6 +339,18 @@ namespace skyline {
...
             return frz::elsa<frz::string>{}(frz::string(view.data(), view.size()), 0);
         }
+
+        /**
+         * @brief Fills an array with random data from a Mersenne Twister pseudo-random generator
+         * @note Currently the generator is not seeded so produced values will be identical between two runs
+         */
+        template<typename T> requires (std::is_integral_v<T>)
+        void GetRandomBytes(std::span<T> bytes) {
+            thread_local std::mt19937_64 generator{};
```
PixelyIon, line 349: I could be wrong but isn't brace initialization pointless for non-POD types like `mt19937_64` where the default constructor would automatically be run

app/src/main/cpp/skyline/services/spl/IRandomInterface.cpp
```diff
@@ -0,0 +1,17 @@
...
+
+#include <common.h>
+
+namespace skyline::service::spl {
+    IRandomInterface::IRandomInterface(const DeviceState &state, ServiceManager &manager) : BaseService(state, manager) {}
+
+    Result IRandomInterface::GetRandomBytes(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
+        auto &outBuf = request.outputBuf.at(0);
+
+        util::GetRandomBytes(outBuf);
```
PixelyIon, line 14:
```suggestion
        auto &outBuf{request.outputBuf.at(0)};

        util::GetRandomBytes(outBuf);
```
or
```suggestion
        util::GetRandomBytes(request.outputBuf.at(0));
```

PixelyIon (changes requested):
Just teensy little nitpicks

app/src/main/cpp/skyline/common.h
bylaws: Also make uuid use GetRandomBytes

app/src/main/cpp/skyline/common.h
```diff
@@ -338,6 +339,18 @@ namespace skyline {
         constexpr std::size_t Hash(std::string_view view) {
             return frz::elsa<frz::string>{}(frz::string(view.data(), view.size()), 0);
         }
+
+        /**
+         * @brief Fills an array with random data from a Mersenne Twister pseudo-random generator
+         * @note Currently the generator is not seeded so produced values will be identical between two runs
+         */
+        template<typename T, typename S = u64> requires (std::is_integral_v<T> && std::is_integral_v<S>)
```
PixelyIon, line 347: Usage of `T` here is going to be problematic since you'll have multiple `std::mt19937_64` generators based on the type. You'd want to move the `generator` outside of the function (in a `util::detail` namespace) so that you only have one generator per-thread that multiple `std::uniform_int_distribution`s can utilize.

app/src/main/cpp/skyline/common.h
```diff
@@ -338,6 +339,18 @@ namespace skyline {
...
+        void GetRandomBytes(std::span<T> bytes, S seed = GetTimeTicks()) {
```
PixelyIon, line 348: Why is the span parameter called `bytes` when it's not a byte-span, I also disagree with the function naming as it's not really a getter as much as it is filling a `span` with random data. I'd recommend something like `FillRandomBytes` rather than `GetRandomBytes`.

PixelyIon (changes requested):
I'm sorry I didn't actually take much of a look at the generator function itself the last time, this should cover the issues with it.

bylaws (approved):
I approve 😸

app/src/main/cpp/skyline/common.h
```diff
@@ -338,6 +339,35 @@ namespace skyline {
         constexpr std::size_t Hash(std::string_view view) {
             return frz::elsa<frz::string>{}(frz::string(view.data(), view.size()), 0);
         }
+
+        /**
+         * @brief Returns the most appropriate integral type and count to represent the given T object
```
PixelyIon, line 344: I'd actually recommend changing this a bit and naming it `PointerFor` (also change `type` to a pointer type) and explicitly stating that it selects the largest possible pointer type for representing an object alongside providing the size of the object in terms of the underlying type.

app/src/main/cpp/skyline/common.h
```diff
@@ -338,6 +339,35 @@ namespace skyline {
...
+        /**
+         * @brief Fills an array with random data from a Mersenne Twister pseudo-random generator
+         * @note The generator is seeded with the the current time in ticks
+         */
+        template<typename T> requires (std::is_integral_v<T>)
+        void FillRandomBytes(std::span<T> in) {
+            std::independent_bits_engine<std::mt19937_64, std::numeric_limits<T>::digits, T> gen(detail::generator);
+
+            std::generate(in.begin(), in.end(), gen);
+        }
```
PixelyIon, line 370: Do one that is `requires (!std::is_integral_v<T>)` with just `void FillRandomBytes(T object)` and calls `FillRandomBytes(std::span(reinterpret_cast<PointerFor<T>::type>(&object), PointerFor<T>::count))` which should handle the case for non-integral types then utilize this for the UUIDs. Alternatively, you can make a `T CreateRandomObject()` which takes nothing but returns an object filled with random bytes of whatever the supplied template type is.

PixelyIon (changes requested):

PixelyIon (approved):
Great job, everything looks perfect!

PR #304, opened 2022-01-07
Compile-time Logger Format String Validation + Async Logger

nickbeth:
This PR accomplishes the following:
1. {fmt} [8.0.0](https://github.com/fmtlib/fmt/releases/tag/8.0.0) enabled compile-time format string checking by default on compilers with C++20 `consteval` support. All format strings were previously wrapped in `fmt::runtime` waiting to comply with the new change. This PR finally makes use of compile-time checking by taking `fmt::format_string` as format string argument type in the `Logger` calls.
2. We noticed logging IO operations took almost as much as shader compilation. This PR offloads IO operations to a dedicated thread, log messages are pushed to a FIFO and will be written to logcat/file by the logger thread.

TODO:
- [ ] Merge `ftx1` branch
- [ ] Refactor all Logger calls to the new macro ones


Newman1515 (commented):
Submit the approve
And 
Submit the change

Sovietfowl:
Your approval for every pull request is very appreciated but you don't need to comment on every one of them.

PR #324, opened 2022-01-13
Settings infrastructure rework

lynxnb:
This PR reworks how settings are shared between Kotlin and native code. We used to pass the file descriptor of the SharedPreferences xml to be parsed by the `Settings` class using PugiXML.
Instead, now settings are kept in a Kotlin class, and an instance is passed to the native side when running a game. Native side then constructs a `Settings` instance from the value of the Kotlin instance accessed via JNI.

This allows the following:
* Native `Settings` class is now platform-independent, and doesn't rely on Android specific files/behavior
* We now have a simple and reliable way to update settings during emulation
* Single settings now also allow adding callbacks that will be run when their value changes


Newman1515 (commented):
Submit the approve
And
Submit the change

PixelyIon:
@Newman1515 Please don't spam this on PRs, it won't fast-track merging. We know what we're doing and will merge when we determine it's a good time to do so. Doing more of these will result in a temporary ban.

app/src/main/cpp/emu_jni.cpp
```diff
@@ -85,8 +84,8 @@ extern "C" JNIEXPORT void Java_emu_skyline_EmulationActivity_executeApplication(
     pthread_setname_np(pthread_self(), "EmuMain");
 
     auto jvmManager{std::make_shared<skyline::JvmManager>(env, instance)};
-    auto settings{std::make_shared<skyline::Settings>(preferenceFd)};
-    close(preferenceFd);
+
+    std::shared_ptr<skyline::Settings> settings{std::make_shared<skyline::AndroidSettings>(env, settingsInstance)};
```
bylaws, line 88: Auto

app/src/main/cpp/skyline/common/settings.h
```diff
@@ -3,23 +3,72 @@
...
+
       public:
-        Logger::LogLevel logLevel; //!< The minimum level that logs need to be for them to be printed
-        std::string username; //!< The name set by the user to be supplied to the guest
-        bool operationMode; //!< If the emulated Switch should be handheld or docked
-        bool forceTripleBuffering; //!< If the presentation engine should always triple buffer even if the swapchain supports double buffering
-        bool disableFrameThrottling; //!< Allow the guest to submit frames without any blocking calls
+        // System
+        Setting<bool> isDocked; //!< If the emulated Switch should be handheld or docked
+        Setting<std::string> usernameValue; //!< The name set by the user to be supplied to the guest
```
bylaws, line 57: userName

app/src/main/cpp/emu_jni.cpp
lynxnb: Auto won't make use of polymorphism though, or is c++ able to upcast automatically?

bylaws (changes requested):
Codes

app/src/main/cpp/skyline/common/settings.h
```diff
@@ -3,23 +3,72 @@
...
+        Setting<language::SystemLanguage> systemLanguage; //!< The system language set by the user
+
+        // Display
+        Setting<bool> forceTripleBuffering; //!< If the presentation engine should always triple buffer even if the swapchain supports double buffering
+        Setting<bool> disableFrameThrottling; //!< Allow the guest to submit frames without any blocking calls
+
+        Settings() {}
```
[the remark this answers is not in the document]
lynxnb: I don't think we want this to be a POD when it's not, I think we should keep the empty ctor.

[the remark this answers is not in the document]
bylaws: The compiler isn't silly, it won't mark a type as trivial just because of a defaulted ctor, defaulted ctor just doesn't prevent it. In this case it's not trivial cause of the virtual member. Defaulting a ctor otherwise gives exact same behaviour while being more uniform

PixelyIon (approved):
LGTM!

app/src/main/cpp/skyline/common/android_settings.h
```diff
@@ -0,0 +1,41 @@
...
+      public:
+        /**
+         * @brief Will construct the underlying KtSettings object in-place
+         */
+        AndroidSettings(JNIEnv *env, jobject settingsInstance) : ktSettings(env, settingsInstance) {
+            Update();
+        }
+
+        /**
+         * @brief A constructor that takes ownership of the passed KtSettings object
```
bylaws, line 27: We know it's a ctor, you can drop that '@note Will take ownership of the passed KtSettings object'

app/src/main/cpp/skyline/common/android_settings.h
```diff
@@ -0,0 +1,41 @@
...
+     * @brief Handles settings on the android platform
+     * @note Lifetime of this class must not exceed the one of the JNIEnv contained inside ktSettings
+     */
+    class AndroidSettings final : public Settings {
+      private:
+        KtSettings ktSettings;
+
+      public:
+        /**
+         * @brief Will construct the underlying KtSettings object in-place
```
bylaws, line 20: @note

app/src/main/cpp/skyline/common/android_settings.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include "settings.h"
+#include "jvm.h"
+
+namespace skyline {
+    /**
+     * @brief Handles settings on the android platform
+     * @note Lifetime of this class must not exceed the one of the JNIEnv contained inside ktSettings
+     */
+    class AndroidSettings final : public Settings {
```
bylaws, line 14: This class should be moved to a cpp file

app/src/main/cpp/skyline/common/settings.h
```diff
@@ -3,23 +3,78 @@
...
+            std::mutex valueMutex;
+            std::mutex callbackMutex;
+
+            /**
+             * @brief Calls all callbacks registered for this setting
+             * @note Locking of the setting value must be handled by the caller
+             */
+            void OnSettingChanged() {
+                std::scoped_lock lock{callbackMutex};
+                for (const Callback &callback : callbacks)
```
bylaws, line 27: const auto &

bylaws (approved):

PR #367, opened 2022-02-04
Ignore empty lines in key files

K0bin:
I wanted to try skyline and it kept rejecting my prod.keys file. Turned out to be caused by an empty line.


PixelyIon (approved):
LGTM!

PR #413, opened 2022-03-04
Minor Vulkan fixes

K0bin:
This is a bunch of minor things I noticed while trying to run Mario Odyssey.

### DIFF
--- a/app/src/main/cpp/skyline/gpu/texture/texture.cpp
+++ b/app/src/main/cpp/skyline/gpu/texture/texture.cpp
@@ -272,7 +272,7 @@ namespace skyline::gpu {
           layerCount(guest->layerCount),
           sampleCount(vk::SampleCountFlagBits::e1) {
         vk::ImageUsageFlags usage{vk::ImageUsageFlagBits::eTransferSrc | vk::ImageUsageFlagBits::eTransferDst | vk::ImageUsageFlagBits::eSampled};
-        if (format->vkAspect & vk::ImageAspectFlagBits::eColor)
+        if ((format->vkAspect & vk::ImageAspectFlagBits::eColor) && !format->IsCompressed())
             usage |= vk::ImageUsageFlagBits::eColorAttachment;
         if (format->vkAspect & (vk::ImageAspectFlagBits::eDepth | vk::ImageAspectFlagBits::eStencil))
             usage |= vk::ImageUsageFlagBits::eDepthStencilAttachment;
--- a/app/src/main/cpp/skyline/gpu/texture/texture.h
+++ b/app/src/main/cpp/skyline/gpu/texture/texture.h
@@ -80,7 +80,7 @@ namespace skyline::gpu {
              * @return The size of the texture in bytes
              */
             constexpr size_t GetSize(u32 width, u32 height, u32 depth = 1) const {
-                return (((width / blockWidth) * (height / blockHeight)) * bpb) * depth;
+                return util::DivideCeil(width, u32{blockWidth}) * util::DivideCeil(height, u32{blockHeight}) * bpb * depth;
             }
 
             constexpr size_t GetSize(Dimensions dimensions) const {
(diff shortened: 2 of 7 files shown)


app/src/main/cpp/skyline/gpu/trait_manager.cpp
```diff
@@ -79,6 +79,12 @@ namespace skyline::gpu {
             enabledFeatures2.unlink<vk::PhysicalDeviceVertexAttributeDivisorFeaturesEXT>();
         }
 
+        if (supportsLastProvokingVertex) {
```
bylaws, line 82: Can drop the {} here, see our style guide

bylaws (changes requested):

app/src/main/cpp/skyline/gpu/trait_manager.cpp
K0bin: Done.

bylaws:
Thanks, will let mark check but otherwise looks good for me :))

app/src/main/cpp/skyline/gpu/trait_manager.cpp
```diff
@@ -79,6 +79,11 @@ namespace skyline::gpu {
             enabledFeatures2.unlink<vk::PhysicalDeviceVertexAttributeDivisorFeaturesEXT>();
         }
 
+        if (supportsLastProvokingVertex)
+            FEAT_SET(vk::PhysicalDeviceProvokingVertexFeaturesEXT, provokingVertexLast, supportsLastProvokingVertex)
+        else
+            enabledFeatures2.unlink<vk::PhysicalDeviceProvokingVertexFeaturesEXT>();
```
PixelyIon, line 85: As `VK_EXT_provoking_vertex` has always had the feature structure, I'd say the `EXT_SET` for this should set a local boolean `hasLastProvokingVertexExt` then this should branch off that variable.

(PS: You can make all local variable naming a bit more consistent with `bool hasCustomBorderColorExt{}, hasShaderAtomicInt64Ext{}, hasShaderFloat16Int8Ext{}, hasShaderDemoteToHelperExt{}, hasProvokingVertexExt{};` while you're at it)

app/src/main/cpp/skyline/gpu/texture/texture.cpp
```diff
@@ -272,7 +272,7 @@ namespace skyline::gpu {
           layerCount(guest->layerCount),
           sampleCount(vk::SampleCountFlagBits::e1) {
         vk::ImageUsageFlags usage{vk::ImageUsageFlagBits::eTransferSrc | vk::ImageUsageFlagBits::eTransferDst | vk::ImageUsageFlagBits::eSampled};
-        if (format->vkAspect & vk::ImageAspectFlagBits::eColor)
+        if (format->vkAspect & vk::ImageAspectFlagBits::eColor && format->blockWidth == 1 && format->blockHeight == 1)
```
PixelyIon, line 275: Would be optimal if a `format->IsCompressed()` function was added to do this check in a more explicit manner.

app/src/main/cpp/skyline/gpu/texture/texture.h
```diff
@@ -80,7 +80,7 @@ namespace skyline::gpu {
              * @return The size of the texture in bytes
              */
             constexpr size_t GetSize(u32 width, u32 height, u32 depth = 1) const {
-                return (((width / blockWidth) * (height / blockHeight)) * bpb) * depth;
+                return ((((width + blockWidth - 1) / blockWidth) * ((height + blockHeight - 1) / blockHeight)) * bpb) * depth;
```
PixelyIon, line 83: I'd say separating this into `util::DivideCeil` would be more neat, place it after `IsWordAligned` in `common/utils.h`. It can just be `util::DivideCeil(width, blockWidth) * util::DivideCeil(height, blockHeight) * bpb * depth` at that point.

PixelyIon (changes requested):
Great job with this, I just have minor suggestions to improve readability/correctness in some cases.

PixelyIon (approved):
LGTM, good work!

PR #435, opened 2022-03-08
Prevent null pointer exception

cbedoy:
I've seen that several parts in the code have been use -!!- in order to have not null instance, the problem here is forcing an instance that actual value is null will be lead a crash.

### DIFF
--- a/app/src/main/java/emu/skyline/AppDialog.kt
+++ b/app/src/main/java/emu/skyline/AppDialog.kt
@@ -72,7 +72,7 @@ class AppDialog : BottomSheetDialogFragment() {
         super.onViewCreated(view, savedInstanceState)
         view.setBackgroundColor(requireContext().getColor(R.color.backgroundColor))
 
-        val missingIcon = ContextCompat.getDrawable(requireActivity(), R.drawable.default_icon)!!.toBitmap(256, 256)
+        val missingIcon = ContextCompat.getDrawable(requireActivity(), R.drawable.default_icon)?.toBitmap(256, 256)
 
         binding.gameIcon.setImageBitmap(item.icon ?: missingIcon)
         binding.gameTitle.text = item.title
--- a/app/src/main/java/emu/skyline/EmulationActivity.kt
+++ b/app/src/main/java/emu/skyline/EmulationActivity.kt
@@ -199,9 +199,9 @@ class EmulationActivity : AppCompatActivity(), SurfaceHolder.Callback, View.OnTo
         if (emulationThread?.isAlive == true) {
             shouldFinish = false
             if (stopEmulation(false))
-                emulationThread!!.join(250)
+                emulationThread?.join(250)
 
-            if (emulationThread!!.isAlive) {
+            if (emulationThread?.isAlive == true) {
                 finishAffinity()
                 startActivity(intent)
                 Runtime.getRuntime().exit(0)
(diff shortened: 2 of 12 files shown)


bylaws:
The current ones are left as they are because it's literally impossible for them to be null and a crash is desirable if they are.

app/src/main/java/emu/skyline/AppDialog.kt
```diff
@@ -72,7 +72,7 @@ class AppDialog : BottomSheetDialogFragment() {
         super.onViewCreated(view, savedInstanceState)
         view.setBackgroundColor(requireContext().getColor(R.color.backgroundColor))
 
-        val missingIcon = ContextCompat.getDrawable(requireActivity(), R.drawable.default_icon)!!.toBitmap(256, 256)
+        val missingIcon = ContextCompat.getDrawable(requireActivity(), R.drawable.default_icon)?.toBitmap(256, 256)
```
lynxnb, line 75: - If the call to `ContextCompat.getDrawable()` fails, we have bigger problems than the image being null. And even if that wasn't the case, there's no fallback image, so the ImageView would be blank.

app/src/main/java/emu/skyline/EmulationActivity.kt
```diff
@@ -211,17 +211,27 @@ class EmulationActivity : AppCompatActivity(), SurfaceHolder.Callback, View.OnTo
...
+                returnFromEmulation()
+            }
 
-        emulationThread = Thread {
-            executeApplication(rom.toString(), romType, romFd.detachFd(), preferenceFd.detachFd(), settings.systemLanguage, applicationContext.filesDir.canonicalPath + "/", applicationInfo.nativeLibraryDir + "/", assets)
-            returnFromEmulation()
+            emulationThread?.start()
         }
-
-        emulationThread!!.start()
```
lynxnb, line 224: This is code that start the emulation thread, which in turns runs the game.
Here you just replaced the not-null assertion operator with a safe calls, but without providing a viable fallback path in case the value is null. The app would do nothing in this case and just hang, this is even worse than crashing.

app/src/main/java/emu/skyline/EmulationActivity.kt
```diff
@@ -263,7 +273,7 @@ class EmulationActivity : AppCompatActivity(), SurfaceHolder.Callback, View.OnTo
             }
         }
 
-        @Suppress("DEPRECATION") val display = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) display!! else windowManager.defaultDisplay
+        @Suppress("DEPRECATION") val display = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) display else windowManager.defaultDisplay
```
lynxnb, line 276: Same issue as the `ContextCompat` case above, if `Context.getDisplay()` is null, we have bigger problems than the NPE crash.

PixelyIon:
While I don't see much point of this commit due to the aforementioned reasons, we do have plenty of stuff on the Kotlin side we want to do and if you want to contribute come over to the [Discord server](https://discord.gg/XnbXNQM).

lynxnb (commented):
As @bylaws already said, a crash is desirable when an `!!` is present, as it would mean the whole Android subsystem has crashed or it isn't working properly. Also since we have to do with Java objects, NPE is something we have to deal with (sadly).
I've highlighted some of the instances where crashing is an acceptable behaviour, but you can apply the same reasoning to the other changes as well.

lynxnb:
Also since you mention "Preventing NPE crashes" in the title, did you experience any crash that is directly related to `!!` being used?

app/src/main/java/emu/skyline/EmulationActivity.kt
AbandonedCart: It might solve the concern to crash more gracefully in such instances, since a fatal crash looks unplanned and it sounds like a crash is an expected outcome.

MCredstoner2004: > It might solve the concern to crash more gracefully in such instances, since a fatal crash looks unplanned and it sounds like a crash is an expected outcome.

It wouldn't be more gracefully, because yes, if this goes wrong, a crash is the expected behavior, and this doesn't even throw a more meaningful exception so it's probably less useful for debugging than a NPE

AbandonedCart: @MCredstoner2004 I think you misunderstood. I was suggesting that the crash is *handled* more gracefully, as it is currently not being handled at all. That doesn't mean *preventing* it, as the PR is attempting.

lynxnb: I see what you mean, but there's really nothing to handle here. If those objects are null where the `!!` are placed, the app should crash as that means unrecoverable failure. Although, the circumstances when this would happen are very rare.

PR #750, opened 2022-03-22
Time to pull some strings

AlstonHollingsworthWaterhouse:
Please do your job trannies, I know you're busy dilating but it can't be so hard making a better emulator than EggNS.

Maybe stop using Twitter all the time, and start developing it's not like you trannies have anything better to do.


MrKing202981:
We're not merging the FTX-1 Branch into Master due to it containing changes deemed unstable. Calling the developers transphobic slurs won't do you any good either, as such I'll be banning you from the repository.

PR #784, opened 2022-03-23
Stub `nfp` service + `ListAddonContent` + `IsVrModeEnabled `

lynxnb:
These services are called in a loop by **The Legend of Zelda: Breath of the Wild** if not implemented.


PixelyIon (approved):
LGTM!

PR #793, opened 2022-03-24
Automatically import key files from search location

lynxnb:
`title.keys` and `prod.keys` files found in the search location folder are now imported automatically (subfolders included). This means it should no longer be necessary to manually import keys if they are kept in the same folder as games.


PixelyIon (approved):
LGTM!

PR #839, opened 2022-03-28
Stub ICommonStateGetter::SetCpuBoostMode

MK73DS:
This makes Metroid Dread boot, although it has very major graphical glitches. (Screenshot is from the game's gallery)
image

### DIFF
--- a/app/src/main/cpp/skyline/services/am/controller/ICommonStateGetter.cpp
+++ b/app/src/main/cpp/skyline/services/am/controller/ICommonStateGetter.cpp
@@ -71,4 +71,19 @@ namespace skyline::service::am {
         }
         return {};
     }
+
+    Result ICommonStateGetter::SetCpuBoostMode(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
+        cpuBoostMode = request.Pop<CpuBoostMode>();
+        switch (cpuBoostMode) {
+            case CpuBoostMode::Normal:
+            case CpuBoostMode::FastLoad:
+            case CpuBoostMode::PowerSaving:
+                Logger::Info("Set CPU boost mode to {}", ToString(cpuBoostMode));
+                return {};
+            default:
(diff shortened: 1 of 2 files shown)


app/src/main/cpp/skyline/services/am/controller/ICommonStateGetter.cpp
```diff
@@ -71,4 +71,9 @@ namespace skyline::service::am {
         }
         return {};
     }
+
+    // Stub implementation of SetCpuBoostMode
```
PixelyIon, line 75: You can remove this comment, we generally don't do this regardless of something being a stub.

app/src/main/cpp/skyline/services/am/controller/ICommonStateGetter.cpp
```diff
@@ -71,4 +71,9 @@ namespace skyline::service::am {
...
+    Result ICommonStateGetter::SetCpuBoostMode(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
+        return {};
```
PixelyIon, line 77: You should check for the value being valid here, probably write a `CpuBoostMode` enum then create `constexpr Result InvalidParameters(128, 506)` in the header under the `result` namespace and return it depending on if the value is valid.

app/src/main/cpp/skyline/services/am/controller/ICommonStateGetter.h
```diff
@@ -95,14 +95,21 @@ namespace skyline::service::am {
          */
         Result GetDefaultDisplayResolution(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response);
 
+        /**
+         * @brief Takes an input u32 #CpuBoostMode and passes it to SetCpuBoostMode.
```
PixelyIon, line 99: That description doesn't make much sense, `Sets the CPU boost mode to the supplied value` would make more sense.

PixelyIon (changes requested):

PixelyIon (approved):
LGTM!

PR #905, opened 2022-04-04
nce/guest: Fix counting error

merryhime:
Fix constants to reflect code change in commit cffbfc80345438c8ffe26b40bf7595e6f99dc8d7.

### DIFF
--- a/app/src/main/cpp/skyline/nce/guest.h
+++ b/app/src/main/cpp/skyline/nce/guest.h
@@ -103,8 +103,8 @@ namespace skyline {
         };
 
         namespace guest {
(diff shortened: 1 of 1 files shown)


bylaws:
hey~ I hate to say this on your first pr... It would have been fine literally two days earlier but mark just started work on supporting thread context svcs and they require us to store all of the register state anyway so this would be kinda redundant to merge.. sorry! Thanks for spotting this anyway though, will remind us not not forget about it in the future

(Also sorry you had to look though the mess that is our old commits, thankfully they're a lot better now after I spent an hour convincing mark :))
(Also we're doing a lot of work on the ftx1 branch now, I recommend you look at that for up to date stuffs if you're curious)

merryhime:
Ah, thanks, will take a look.

PR #940, opened 2022-04-09
Minor readability change

aecarvalho:
Just a minor tweak to make it more concise

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 ---
 
 ### Contact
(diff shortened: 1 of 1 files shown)


PixelyIon:
I feel like this is a bit too pedantic and unnecessary to be done on its own, a rewording can be included in the `readme` branch instead but I do think it's somewhat unnecessary still.

aecarvalho:
Doesn't mean it's necessarily erroneous either, but okay.

PR #946, opened 2022-04-09
Fix IApplicationFunctions::GetDisplayVersion id

MK73DS:
According to [SwitchBrew](https://switchbrew.org/wiki/Applet_Manager_services#IApplicationFunctions), it should be 0x17 and not 0x16


lynxnb (approved):
LGTM

PR #1022, opened 2022-04-14
Make on-screen buttons more visible and change their position

dimaxdqwerty:
Button symbols have same color as buttons, so to changing color to another make buttons more visible. Also position of buttons is not great, so this is can be improved.


haradox0:
and if possible
moving dpad together like aethersx2 ,same for abxy 🙏🏻

xxRaiyanxxYT:
Oh and also gib options to adjust opacity of the buttons since its barely visible in some games xD

Newman1515 (commented):
Submit the approve 
  

Submit the review

PixelyIon:
We have other plans for this at the moment which will likely result in this commit being superseded.

aungko122:
s

Ronzoo:
Add GTA SA remastered

PR #1136, opened 2022-04-29
feat: added italian translation

Thepowa753:
Just added this translate (useful also as template for other translations)


lynxnb:
We appreciate the effort you put into this, but this is not how we intend to do translations. We'll likely be using Transifex or something of that sort, and then automatically pull translations from there.

I would suggest you keep these translations around for when we'll properly setup translations, so that you can contribute them at the right time.

I will be closing this PR.

PR #1300, opened 2022-05-14
Nvdrv gpu misc

MenU15:
Exnyos 990 
Itried so many times to open up any game but it stil getting out from the and any such of game please help


Roydpeter:
Mali GPUs don't work as of now. Give it a few months or years. Your phone has a Mali GPU. A weak one at that. All Mali GPUs are weak when it comes to emulation

PR #1386, opened 2022-05-27
Add HID touch attributes and touch index

german77:
Adds missing parameter TouchAttribute  and emulates correctly the touch point index. Both changes are necessary on Voez to keep track of each finger.

### DIFF
--- a/app/src/main/cpp/skyline/input/sections/TouchScreen.h
+++ b/app/src/main/cpp/skyline/input/sections/TouchScreen.h
@@ -7,12 +7,26 @@
 
 namespace skyline::input {
     /**
+     * @brief Indicates if touch point has started or ended
+     * @url https://switchbrew.org/wiki/HID_services#TouchAttribute
+     */
+    union TouchAttribute {
+        u32 raw{};
+        struct {
+            bool start : 1;
+            bool end : 1;
+        };
+    };
+    static_assert(sizeof(TouchAttribute) == 0x4);
+
+    /**
      * @brief A descriptor for a single point on the touch screen
      * @url https://switchbrew.org/wiki/HID_Shared_Memory#TouchScreenStateData
      */
     struct TouchScreenStateData {
         u64 timestamp; //!< The timestamp in samples
-        u32 _pad0_;
+
+        TouchAttribute attribute;
 
         u32 index; //!< The index of this touch
 
--- a/app/src/main/cpp/skyline/input/touch.cpp
+++ b/app/src/main/cpp/skyline/input/touch.cpp
@@ -15,13 +15,14 @@ namespace skyline::input {
         }
     }
 
-    void TouchManager::SetState(const span<TouchScreenPoint> &points) {
+    void TouchManager::SetState(const span<TouchScreenPoint> touchPoints) {
         if (!activated)
             return;
 
         const auto &lastEntry{section.entries[section.header.currentEntry]};
         auto entryIndex{(section.header.currentEntry != constant::HidEntryCount - 1) ? section.header.currentEntry + 1 : 0};
         auto &entry{section.entries[entryIndex]};
+        auto points{touchPoints.subspan(0, std::min(touchPoints.size(), entry.data.size()))};
         entry.globalTimestamp = lastEntry.globalTimestamp + 1;
         entry.localTimestamp = lastEntry.localTimestamp + 1;
         entry.touchCount = points.size();
@@ -29,13 +30,18 @@ namespace skyline::input {
         for (size_t i{}; i < points.size(); i++) {
             const auto &host{points[i]};
             auto &guest{entry.data[i]};
-            guest.index = static_cast<u32>(i);
+            guest.attribute.raw = static_cast<u32>(host.attribute);
+            guest.index = static_cast<u32>(host.id);
             guest.positionX = static_cast<u32>(host.x);
             guest.positionY = static_cast<u32>(host.y);
             guest.minorAxis = static_cast<u32>(host.minor);
             guest.majorAxis = static_cast<u32>(host.major);
             guest.angle = host.angle;
         }
+
+        // Clear unused touch points
+        for (size_t i{points.size()}; i < entry.data.size(); i++)
+            entry.data[i] = {};
 
         section.header.timestamp = util::GetTimeTicks();
         section.header.entryCount = std::min(static_cast<u8>(section.header.entryCount + 1), constant::HidEntryCount);
--- a/app/src/main/cpp/skyline/input/touch.cpp
+++ b/app/src/main/cpp/skyline/input/touch.cpp
@@ -15,14 +15,14 @@ namespace skyline::input {
         }
     }
 
-    void TouchManager::SetState(const span<TouchScreenPoint> touchPoints) {
+    void TouchManager::SetState(span<TouchScreenPoint> touchPoints) {
         if (!activated)
             return;
 
         const auto &lastEntry{section.entries[section.header.currentEntry]};
         auto entryIndex{(section.header.currentEntry != constant::HidEntryCount - 1) ? section.header.currentEntry + 1 : 0};
         auto &entry{section.entries[entryIndex]};
-        auto points{touchPoints.subspan(0, std::min(touchPoints.size(), entry.data.size()))};
+        touchPoints = touchPoints.first(std::min(touchPoints.size(), entry.data.size()));
         entry.globalTimestamp = lastEntry.globalTimestamp + 1;
         entry.localTimestamp = lastEntry.localTimestamp + 1;
         entry.touchCount = points.size();
--- a/app/src/main/cpp/skyline/input/touch.cpp
+++ b/app/src/main/cpp/skyline/input/touch.cpp
@@ -25,10 +25,10 @@ namespace skyline::input {
         touchPoints = touchPoints.first(std::min(touchPoints.size(), entry.data.size()));
         entry.globalTimestamp = lastEntry.globalTimestamp + 1;
         entry.localTimestamp = lastEntry.localTimestamp + 1;
-        entry.touchCount = points.size();
+        entry.touchCount = touchPoints.size();
 
-        for (size_t i{}; i < points.size(); i++) {
-            const auto &host{points[i]};
+        for (size_t i{}; i < touchPoints.size(); i++) {
+            const auto &host{touchPoints[i]};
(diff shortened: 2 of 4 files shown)


app/src/main/cpp/skyline/input/touch.cpp
```diff
@@ -26,17 +26,23 @@ namespace skyline::input {
         entry.localTimestamp = lastEntry.localTimestamp + 1;
         entry.touchCount = points.size();
 
-        for (size_t i{}; i < points.size(); i++) {
+        for (size_t i{}; i < points.size() && i < MaxTouchPoints; i++) {
```
bylaws, line 29: Is it fine if points.size() is greater than 16? Cause if it is then touchcount will be larger than the number of populated points

app/src/main/cpp/skyline/input/touch.cpp
```diff
@@ -26,17 +26,23 @@ namespace skyline::input {
...
+            guest.index = static_cast<u32>(host.id);
             guest.positionX = static_cast<u32>(host.x);
             guest.positionY = static_cast<u32>(host.y);
             guest.minorAxis = static_cast<u32>(host.minor);
             guest.majorAxis = static_cast<u32>(host.major);
             guest.angle = host.angle;
         }
 
+        // Clear unused touch points
+        for (size_t i{points.size()}; i < MaxTouchPoints; i++) {
```
PixelyIon, line 42: You can remove the `{}` braces here as it's a single line body.

app/src/main/java/emu/skyline/EmulationActivity.kt
```diff
@@ -477,6 +477,16 @@ class EmulationActivity : AppCompatActivity(), SurfaceHolder.Callback, View.OnTo
             val x = 0f.coerceAtLeast(pointer.x * 1280 / view.width).toInt()
             val y = 0f.coerceAtLeast(pointer.y * 720 / view.height).toInt()
 
+            var attribute = 0
```
PixelyIon, line 480: Do the if statement as if-else and use Kotlin inline if assignment syntax: `val attribute = if (action == ...) 1 else if (...) 2 else 0` and maybe even skip assigning the variable and directly assign it to the array.

app/src/main/cpp/skyline/input/sections/TouchScreen.h
```diff
@@ -6,13 +6,29 @@
...
+
     /**
      * @brief A descriptor for a single point on the touch screen
      * @url https://switchbrew.org/wiki/HID_Shared_Memory#TouchScreenStateData
      */
     struct TouchScreenStateData {
         u64 timestamp; //!< The timestamp in samples
-        u32 _pad0_;
+
+        TouchAttribute attribute; //!< The attribute of this thouch
```
PixelyIon, line 31: touch* but I will add that this is somewhat older code and we have stopped doing comments like this where reading the code would imply the same thing faster so you can just entirely remove this comment really.

PixelyIon (changes requested):
Great job, just have a few nits.

bylaws (changes requested):

app/src/main/cpp/skyline/input/touch.cpp
german77: It's not fine if it's bigger since you overflow the array. Although it's really hard to put 16 fingers on a touch screen even if you got 3 hands.

I can add a guarantee from the UI side so this can never happen here

bylaws: Ah okay probably best to just do it at start of function, just do points = points.subspan(0, STD::min(points.size(),  entry.data.size()))

Oh also can you make it take span directly rather than const ref, not sure how that snook past review

PixelyIon: @bylaws It's older code, I'm not sure if we were even using spans much back then.

app/src/main/cpp/skyline/input/touch.cpp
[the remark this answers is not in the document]
PixelyIon: @bylaws Any idea if this can be replaced by magic C++ ranges stuff?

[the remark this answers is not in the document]
bylaws: Not really since it only covers a subspan and a solution like `ranges::generate(span(data).subspan(points.size()), []{return TouchPoint{};})` at least to me seems uglier

app/src/main/java/emu/skyline/EmulationActivity.kt
german77: I don't think is a good idea to avoid assigning the variable since it looses the meaning on the array.

app/src/main/cpp/skyline/input/touch.cpp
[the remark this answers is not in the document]
PixelyIon: Ah well, I agree.

app/src/main/java/emu/skyline/EmulationActivity.kt
AbandonedCart: As Android Studio will warn you, more than two conditions shouldn't be using an inline if else, but instead be using [when](https://www.baeldung.com/kotlin/when#1-when-as-an-expression)

app/src/main/cpp/skyline/input/touch.cpp
```diff
@@ -15,28 +15,34 @@ namespace skyline::input {
...
 
-    void TouchManager::SetState(const span<TouchScreenPoint> &points) {
+    void TouchManager::SetState(const span<TouchScreenPoint> touchPoints) {
         if (!activated)
             return;
 
         const auto &lastEntry{section.entries[section.header.currentEntry]};
         auto entryIndex{(section.header.currentEntry != constant::HidEntryCount - 1) ? section.header.currentEntry + 1 : 0};
         auto &entry{section.entries[entryIndex]};
+        const auto points = touchPoints.subspan(0, std::min(touchPoints.size(), entry.data.size()));
```
bylaws, line 25:
```suggestion
        auto points{touchPoints.subspan(0, std::min(touchPoints.size(), entry.data.size()))};
```

bylaws (changes requested):
One fix but aside from that lgtm

app/src/main/java/emu/skyline/EmulationActivity.kt
german77: Much better looking alternative. Thank you for the suggestion.

app/src/main/cpp/skyline/input/touch.cpp
```diff
@@ -15,28 +15,34 @@ namespace skyline::input {
...
 
-    void TouchManager::SetState(const span<TouchScreenPoint> &points) {
+    void TouchManager::SetState(const span<TouchScreenPoint> touchPoints) {
         if (!activated)
             return;
 
         const auto &lastEntry{section.entries[section.header.currentEntry]};
         auto entryIndex{(section.header.currentEntry != constant::HidEntryCount - 1) ? section.header.currentEntry + 1 : 0};
         auto &entry{section.entries[entryIndex]};
+        auto points{touchPoints.subspan(0, std::min(touchPoints.size(), entry.data.size()))};
```
PixelyIon, line 25: I'd say leave the parameter name as `points` and do `points = points.first(std::min(touchPoints.size(), entry.data.size()); // Curb the amount of points to what can fit inside an entry`.

PixelyIon (changes requested):
I *really* hope this isn't annoying and we generally don't have last minute nits like this but we couldn't catch earlier due to it being a result of a CR fix. I can approve after this change.

app/src/main/cpp/skyline/input/touch.cpp
bylaws: Ehh I like how it is, changing parameters is kinda confusing like that

PixelyIon: `touchPoints` and `points` are not descriptive names or substantially different objects, I feel the semantics of curbing the range make far more sense.

bylaws: I suppose a better variable name then? inputTouchPoints and touchPoints

PixelyIon:
One last thing is the commit name, it doesn't match Skyline's naming convention and would stick out. Removing the part at the start and leaving it as `Add touch attributes and touch index` is fine.

app/src/main/cpp/skyline/input/touch.cpp
PixelyIon: I still prefer the semantics of using `first` as it's already a span. I'd agree if the initial object was a vector but spans already have the semantics of being views into a potentially larger range.

bylaws: first?

PixelyIon: [`span::first`](https://en.cppreference.com/w/cpp/container/span/first) and assigning to the variable as it's clearly an operation that's curbing the range of a `span` and does directly play into the `span`'s design of being an arbitrary view.

bylaws: Ooo didn't know first existed

app/src/main/cpp/skyline/input/touch.cpp
```diff
@@ -15,28 +15,34 @@ namespace skyline::input {
         }
     }
 
-    void TouchManager::SetState(const span<TouchScreenPoint> &points) {
+    void TouchManager::SetState(const span<TouchScreenPoint> touchPoints) {
```
bylaws, line 18:
```suggestion
    void TouchManager::SetState(span<TouchScreenPoint> touchPoints) {
```
We don't use const generally

app/src/main/cpp/skyline/input/touch.cpp
bylaws, line 25:
```suggestion
        touchPoints = touchPoints.first(std::min(touchPoints.size(), entry.data.size()));
```

bylaws:
Also for
> On a side note. I saw that on touch release the fingers where ignored completely. Was something broken when this happened?

I think mark just assumed that that would happen, rather than it being for anything specific

app/src/main/cpp/skyline/input/touch.cpp
PixelyIon: You've stolen my amazing CR comment, I can't believe this.

PR #1402, opened 2022-05-28
Add DocumentProvider for accessing Skyline's files

Vixeliz:
I'm not used to kotlin so a lot of the code is very messy but it works. I can try to fix anything you may have complaints about again sorry about code quality. Seems to work fine though for all the basic things (reading files, creating files, creating folders, moving items, copying items, etc)

### DIFF
--- a/app/src/main/java/emu/skyline/SkylineApplication.kt
+++ b/app/src/main/java/emu/skyline/SkylineApplication.kt
@@ -20,8 +20,18 @@ fun Context.getPublicFilesDir() : File = getExternalFilesDir(null) ?: filesDir
 class SkylineApplication : Application() {
     private external fun initializeLog(appFilesPath : String, logLevel : Int)
 
+    init {
+        instance = this
+    }
+
+    companion object {
+        lateinit var instance : SkylineApplication
+            private set
+    }
+
     override fun onCreate() {
         super.onCreate()
+        instance = this
         System.loadLibrary("skyline")
 
         val appFilesPath = applicationContext.getPublicFilesDir().canonicalPath
--- a/app/src/main/java/emu/skyline/provider/DocumentsProvider.kt
+++ b/app/src/main/java/emu/skyline/provider/DocumentsProvider.kt
@@ -1,0 +1,265 @@
+/*
+ * SPDX-License-Identifier: MPL-2.0
+ * Copyright © 2022 Skyline Team and Contributors (https://github.com/skyline-emu/)
+ */
+
+package emu.skyline.provider
+
+import android.database.Cursor
+import android.database.MatrixCursor
+import android.os.CancellationSignal
+import android.os.ParcelFileDescriptor
+import android.provider.DocumentsContract
+import android.provider.DocumentsProvider
+import android.webkit.MimeTypeMap
+import emu.skyline.R
+import emu.skyline.SkylineApplication
+import emu.skyline.getPublicFilesDir
+import java.io.*
+
+class DocumentsProvider : DocumentsProvider() {
+    private val baseDirectory = File(SkylineApplication.instance.getPublicFilesDir().canonicalPath)
+
+    companion object {
+        private val DEFAULT_ROOT_PROJECTION : Array<String> = arrayOf(
+            DocumentsContract.Root.COLUMN_ROOT_ID,
+            DocumentsContract.Root.COLUMN_MIME_TYPES,
+            DocumentsContract.Root.COLUMN_FLAGS,
+            DocumentsContract.Root.COLUMN_ICON,
+            DocumentsContract.Root.COLUMN_TITLE,
+            DocumentsContract.Root.COLUMN_SUMMARY,
+            DocumentsContract.Root.COLUMN_DOCUMENT_ID,
+            DocumentsContract.Root.COLUMN_AVAILABLE_BYTES
+        )
+
+        private val DEFAULT_DOCUMENT_PROJECTION : Array<String> = arrayOf(
+            DocumentsContract.Document.COLUMN_DOCUMENT_ID,
+            DocumentsContract.Document.COLUMN_MIME_TYPE,
+            DocumentsContract.Document.COLUMN_DISPLAY_NAME,
+            DocumentsContract.Document.COLUMN_LAST_MODIFIED,
+            DocumentsContract.Document.COLUMN_FLAGS,
+            DocumentsContract.Document.COLUMN_SIZE
+        )
+    }
+
+    override fun onCreate() : Boolean {
+        return true
+    }
+
+    /**
+     * @return The [File] that corresponds to the document ID supplied by [getDocumentId]
+     */
+    private fun getFile(documentId : String) : File {
+        val file = File(documentId)
+        if (!file.exists()) throw FileNotFoundException(file.absolutePath + " not found")
+        return file
+    }
+
+    /**
+     * @return A unique ID for the provided [File]
+     */
+    private fun getDocumentId(file : File) : Any? {
+        return file.absolutePath
+    }
+
+    override fun queryRoots(projection : Array<out String>?) : Cursor {
+        val cursor = MatrixCursor(projection ?: DEFAULT_ROOT_PROJECTION)
+        val applicationTitle = SkylineApplication.instance.applicationInfo.loadLabel(SkylineApplication.instance.packageManager).toString()
+        val baseDirectoryDocumentId = getDocumentId(baseDirectory)
+
+        cursor.newRow().apply {
+            add(DocumentsContract.Root.COLUMN_ROOT_ID, baseDirectoryDocumentId)
+            add(DocumentsContract.Root.COLUMN_SUMMARY, null)
+            add(DocumentsContract.Root.COLUMN_FLAGS, DocumentsContract.Root.FLAG_SUPPORTS_CREATE or DocumentsContract.Root.FLAG_SUPPORTS_IS_CHILD)
+            add(DocumentsContract.Root.COLUMN_TITLE, applicationTitle)
+            add(DocumentsContract.Root.COLUMN_DOCUMENT_ID, baseDirectoryDocumentId)
+            add(DocumentsContract.Root.COLUMN_MIME_TYPES, "*/*")
+            add(DocumentsContract.Root.COLUMN_AVAILABLE_BYTES, baseDirectory.freeSpace)
+            add(DocumentsContract.Root.COLUMN_ICON, R.drawable.logo_skyline)
+        }
+
+        return cursor
+    }
+
+    override fun queryDocument(documentId : String?, projection : Array<out String>?) : Cursor {
+        val cursor = MatrixCursor(projection ?: DEFAULT_DOCUMENT_PROJECTION)
+        return includeFile(cursor, documentId, null)
+    }
+
+    override fun isChildDocument(parentDocumentId : String?, documentId : String?) : Boolean {
+        return documentId?.startsWith(parentDocumentId!!) ?: false
+    }
+
+    override fun createDocument(parentDocumentId : String?, mimeType : String?, displayName : String) : String? {
+        var newFile = File(parentDocumentId, displayName)
+        var noConflictId = 1 // Makes sure two files don't have the same name by adding a number to the end
+        while (newFile.exists())
+            newFile = File(parentDocumentId, "$displayName (${noConflictId++})")
+
+        try {
+            if (DocumentsContract.Document.MIME_TYPE_DIR == mimeType) {
+                if (!newFile.mkdir())
+                    throw IOException("Failed to create directory")
+            } else {
+                if (!newFile.createNewFile())
+                    throw IOException("Failed to create file")
+            }
+        } catch (e : IOException) {
+            throw FileNotFoundException("Couldn't create document '${newFile.path}': ${e.message}")
+        }
+
+        return newFile.path
+    }
+
+    override fun deleteDocument(documentId : String?) {
+        val file = getFile(documentId!!)
+        if (!file.delete())
+            throw FileNotFoundException("Couldn't delete document with ID '$documentId'")
+    }
+
+    override fun removeDocument(documentId : String, parentDocumentId : String?) {
+        val parent = getFile(parentDocumentId!!)
+        val file = getFile(documentId)
+
+        var doesFileParentMatch = false
+        val fileParent = file.parentFile
+
+        if (fileParent == null || fileParent.equals(parent))
+            doesFileParentMatch = true
+        if (parent == file || doesFileParentMatch) {
+            if (!file.delete())
+                throw FileNotFoundException("Couldn't delete document with ID '$documentId'")
+        } else {
+            throw FileNotFoundException("Couldn't delete document with ID '$documentId'")
+        }
+    }
+
+    override fun renameDocument(documentId : String?, displayName : String?) : String? {
+        if (displayName == null)
+            throw FileNotFoundException("Couldn't rename document '$documentId' as the new name is null")
+
+        val sourceFile = getFile(documentId!!)
+        val sourceParentFile = sourceFile.parentFile ?: throw FileNotFoundException("Couldn't rename document '$documentId' as it has no parent")
+        val destFile = File(sourceParentFile.path, displayName)
+
+        try {
+            if (!sourceFile.renameTo(destFile))
+                throw FileNotFoundException("Couldn't rename document from '${sourceFile.name}' to '${destFile.name}'")
+        } catch (e : Exception) {
+            throw FileNotFoundException("Couldn't rename document from '${sourceFile.name}' to '${destFile.name}': ${e.message}")
+        }
+        return getDocumentId(destFile).toString()
+    }
+
+    private fun copyDocument(
+        sourceDocumentId : String, sourceParentDocumentId : String,
+        targetParentDocumentId : String?
+    ) : String? {
+        if (!isChildDocument(sourceParentDocumentId, sourceDocumentId))
+            throw FileNotFoundException("Couldn't copy document '$sourceDocumentId' as its parent is not '$sourceParentDocumentId'")
+
+        return copyDocument(sourceDocumentId, targetParentDocumentId)
+    }
+
+    override fun copyDocument(sourceDocumentId : String, targetParentDocumentId : String?) : String? {
+        val parent = getFile(targetParentDocumentId!!)
+        val oldFile = getFile(sourceDocumentId)
+        val newFile = File(parent.path, oldFile.name)
+        try {
+            if (!(newFile.createNewFile() && newFile.setWritable(true) && newFile.setReadable(true)))
+                throw IOException("Couldn't create new file")
+
+            FileInputStream(oldFile).use { inStream ->
+                FileOutputStream(newFile).use { outStream ->
+                    inStream.copyTo(outStream)
+                }
+            }
+        } catch (e : IOException) {
+            throw FileNotFoundException("Couldn't copy document '$sourceDocumentId': ${e.message}")
+        }
+
+        return getDocumentId(newFile).toString()
+    }
+
+    override fun moveDocument(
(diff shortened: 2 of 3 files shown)


Vixeliz:
Ah so sorry forgot to commit the actual provider class

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+        return result
+    }
+
+    override fun isChildDocument(parentDocumentId : String?, documentId : String?) : Boolean {
+        return documentId?.startsWith(parentDocumentId!!) ?: false
+    }
+
+    override fun createDocument(parentDocumentId : String?, mimeType : String?, displayName : String) : String? {
+        var newFile = File(parentDocumentId, displayName)
+        var noConflictId = 2
```
PixelyIon, line 84: It is fairly confusing what `noConflictId` is, if this is required at all I'd recommend adding a simple inline comment explaining what it is.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+        while (newFile.exists()) {
+            newFile = File(parentDocumentId, displayName + " (" + noConflictId++ + ")")
+        }
+        try {
```
PixelyIon, line 88: I'd say there should be newlines before the `try` and after the termination of the `catch` (last `}` in this function) to make this a lot more readable.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+            val succeeded : Boolean = if (DocumentsContract.Document.MIME_TYPE_DIR == mimeType) {
+                newFile.mkdir()
+            } else {
+                newFile.createNewFile()
+            }
+            if (!succeeded) {
+                throw FileNotFoundException("Failed to create document with id " + newFile.path)
+            }
```
PixelyIon, line 96: These are all control flow statements with a single-line body, we omit the `{}` braces in this case.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+            throw FileNotFoundException("Failed to create document with id " + newFile.path)
+        }
+        return newFile.path
+    }
+
+    override fun deleteDocument(documentId : String?) {
+        val file = getFileForDocId(documentId!!)
+        if (!file.delete()) {
+            throw FileNotFoundException("Couldn't delete document with documentId $documentId")
+        }
```
PixelyIon, line 107: Single-line control flow body, brackets aren't necessary. I'll let you go through the whole file to catch this as it'd likely be faster than me pointing out every instance.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+            }
+        }
+        return "application/octect-stream"
+    }
+
+    private fun getFileForDocId(documentId : String) : File {
+        val f = File(documentId)
+        if (!f.exists()) throw FileNotFoundException(f.absolutePath + " not found")
+        return f
+    }
```
PixelyIon, line 289: You'll want to move this right up to where `getDocumentId` is and rename it to `getFile` and add a comment `@return The [File] that corresponds to the document ID supplied by [getDocumentId]`. Additionally, expand `f` to `file` as that's more consistent with the rest of the file and use string templates for the exception.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+            newFile = File(parentDocumentId, displayName + " (" + noConflictId++ + ")")
+        }
+        try {
+            val succeeded : Boolean = if (DocumentsContract.Document.MIME_TYPE_DIR == mimeType) {
+                newFile.mkdir()
+            } else {
+                newFile.createNewFile()
+            }
+            if (!succeeded) {
+                throw FileNotFoundException("Failed to create document with id " + newFile.path)
```
PixelyIon, line 95: `id` should be consistently capitalized in strings as `ID` due to being an abbreviation. Additionally, use template strings here, I'll let you pick up all instances of both these issues throughout the file. I'd also say this can be shortened to `Failed to create document '${newFile.path}'`, do adopt this on further invocations too.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+        } catch (e : IOException) {
+            throw FileNotFoundException("Failed to create document with id " + newFile.path)
+        }
+        return newFile.path
+    }
+
+    override fun deleteDocument(documentId : String?) {
+        val file = getFileForDocId(documentId!!)
+        if (!file.delete()) {
+            throw FileNotFoundException("Couldn't delete document with documentId $documentId")
```
PixelyIon, line 106: `Couldn't delete document '$documentId'`

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    override fun renameDocument(documentId : String?, displayName : String?) : String? {
+
+        if (displayName == null) {
+            throw FileNotFoundException("Couldn't rename document, new name is null")
+        }
+
+        val sourceFile = getFileForDocId(documentId!!)
+        val sourceParentFile = sourceFile.parentFile ?: throw FileNotFoundException("Failed to rename document. File has no parent.")
```
PixelyIon, line 140: `Couldn't rename document '$documentId' as it has no parent`

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+        val destFile = File(sourceParentFile.path, displayName)
+
+        try {
+            val renameSucceeded = sourceFile.renameTo(destFile)
+            if (!renameSucceeded) {
+                throw FileNotFoundException("Failed to rename document. Renamed failed.")
+            }
+        } catch (e : Exception) {
+            throw FileNotFoundException("Couldn't rename document. Error: " + e.message)
```
PixelyIon, line 149: We generally omit things like `Error` in this case as it's somewhat implied, it'd just be `Couldn't rename document from '$originalName' to '$newName': ${e.message}`.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+            } else {
+                throw FileNotFoundException("Failed to delete document with id $documentId")
+            }
+        } else {
+            throw FileNotFoundException("Failed to delete document with id $documentId")
+        }
+    }
+
+    override fun renameDocument(documentId : String?, displayName : String?) : String? {
+
```
PixelyIon, line 134: Useless newline.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+        val sourceFile = getFileForDocId(documentId!!)
+        val sourceParentFile = sourceFile.parentFile ?: throw FileNotFoundException("Failed to rename document. File has no parent.")
+        val destFile = File(sourceParentFile.path, displayName)
+
+        try {
+            val renameSucceeded = sourceFile.renameTo(destFile)
+            if (!renameSucceeded) {
+                throw FileNotFoundException("Failed to rename document. Renamed failed.")
+            }
```
PixelyIon, line 147: Just move the call inside the `if` statement, I will say that it is annoying that this exception is thrown and caught by the outer handler even though I get why.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+        } else {
+            throw FileNotFoundException("Failed to delete document with id $documentId")
+        }
+    }
+
+    override fun renameDocument(documentId : String?, displayName : String?) : String? {
+
+        if (displayName == null) {
+            throw FileNotFoundException("Couldn't rename document, new name is null")
```
PixelyIon, line 136: `Couldn't rename document '$documentId' as the new name is null`

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+
+    private fun copyDocument(
+        sourceDocumentId : String, sourceParentDocumentId : String,
+        targetParentDocumentId : String?
+    ) : String? {
+        if (!isChildDocument(sourceParentDocumentId, sourceDocumentId)) {
+            throw FileNotFoundException(
+                "Failed to copy document with id " +
+                        sourceDocumentId + ". Parent is not: " + sourceParentDocumentId
+            )
```
PixelyIon, line 162: `Failed to copy document '$sourceDocumentId' as its parent is not '$sourceParentDocumentId'`

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
PixelyIon, line 106: You should figure out a consistent convention of using either `Couldn't` or `Failed to` and stick to one of them for all exceptions.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+        }
+        return copyDocument(sourceDocumentId, targetParentDocumentId)
+    }
+
+    override fun copyDocument(sourceDocumentId : String, targetParentDocumentId : String?) : String? {
+
```
PixelyIon, line 168: Useless newline.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+        val parent = getFileForDocId(targetParentDocumentId!!)
+        val oldFile = getFileForDocId(sourceDocumentId)
+        val newFile = File(parent.path, oldFile.name)
+        try {
+
```
PixelyIon, line 173: Useless newline.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+                if (newFile.setWritable(true) && newFile.setReadable(true)) {
+                    wasNewFileCreated = true
+                }
+            }
+            if (!wasNewFileCreated) {
+                throw FileNotFoundException(
+                    "Couldn't copy document " + sourceDocumentId +
+                            ". Could not create new file."
+                )
+            }
```
PixelyIon, line 185: Just do `if (!(newFile.createNewFile() && newFile.setWritable(true) && newFile.setReadable(true)))`.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+                    var len : Int
+                    while ((inStream.read(buf).also { len = it }) > 0) {
+                        outStream.write(buf, 0, len)
+                    }
+                }
+            }
+        } catch (e : IOException) {
+            throw FileNotFoundException(
+                ("Failed to copy document: " + sourceDocumentId +
+                        ". " + e.message)
```
PixelyIon, line 198: `Failed to copy document '$sourceDocumentId': ${e.message}`

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+            }
+            if (!wasNewFileCreated) {
+                throw FileNotFoundException(
+                    "Couldn't copy document " + sourceDocumentId +
+                            ". Could not create new file."
+                )
+            }
+            FileInputStream(oldFile).use { inStream ->
+                FileOutputStream(newFile).use { outStream ->
+                    val buf : ByteArray = ByteArray(4096)
```
PixelyIon, line 188: You don't need `: ByteArray` if you're assigning a `ByteArray` to it, additionally there should be a comment here denoting that you're copying 4 KiB of data at a time. Albeit, I do find it concerning that you need to do a copy in this manner, does `File` not have a method like `File#copy`?

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+                    var len : Int
```
PixelyIon, line 189: `size` rather than `len`.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+                    }
+                }
+            }
+        } catch (e : IOException) {
+            throw FileNotFoundException(
+                ("Failed to copy document: " + sourceDocumentId +
+                        ". " + e.message)
+            )
+        }
+        return getDocIdForFile(newFile).toString()
```
PixelyIon, line 201: A newline before this really, it looks far less cluttered with one.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    override fun moveDocument(
+        sourceDocumentId : String, sourceParentDocumentId : String?,
+        targetParentDocumentId : String?
+    ) : String? {
+        try {
+
```
PixelyIon, line 209: Useless newline.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+            val newDocumentId = copyDocument(
+                sourceDocumentId, sourceParentDocumentId!!,
+                targetParentDocumentId
+            )
+
+            removeDocument(sourceDocumentId, sourceParentDocumentId)
+            return newDocumentId
+        } catch (e : FileNotFoundException) {
+            throw FileNotFoundException("Couldn't move document $sourceDocumentId")
```
PixelyIon, line 218: You'd want `'` around the document ID here.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+
+    private fun includeFile(result : MatrixCursor, documentId : String?, file : File?) : MatrixCursor {
+        var localDocumentId = documentId
+        var localFile = file
+
+        if (localDocumentId == null) {
+            localDocumentId = localFile?.let { getDocIdForFile(it) } as String?
+        } else if (localFile == null) {
+            localFile = getFileForDocId(localDocumentId)
+        }
```
PixelyIon, line 230: You can make this `var localDocumentId = documentId ?: file?.let { getDocIdForFile(it) } as String?` and `var localFile = file ?: getFileForDocId(documentId)`.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+
+        var flags = 0
+        if (localFile == null) {
```
PixelyIon, line 233: You should structure this as `if (localFile != null) ... else ...` as it puts the success condition above.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+            throw FileNotFoundException()
+        } else {
+            if (localFile.isDirectory) {
+                if (localFile.isDirectory && localFile.canWrite()) {
+                    flags = flags or DocumentsContract.Document.FLAG_DIR_SUPPORTS_CREATE
+                }
```
PixelyIon, line 239: Any reason to recheck `localFile.isDirectory` here? Additionally, do remove the `{}` braces here.

app/src/main/java/emu/skyline/provider/SkylineDocumentsProvider.kt
```diff
@@ -0,0 +1,307 @@
...
+            )
+
+            removeDocument(sourceDocumentId, sourceParentDocumentId)
+            return newDocumentId
+        } catch (e : FileNotFoundException) {
+            throw FileNotFoundException("Couldn't move document $sourceDocumentId")
+        }
+    }
+
+    private fun includeFile(result : MatrixCursor, documentId : String?, file : File?) : MatrixCursor {
```
PixelyIon, line 222: `result` -> `cursor`

PixelyIon (changes requested):
You've done a pretty good job, I just have a few style and code-quality related nits.

app/src/main/java/emu/skyline/SkylineApplication.kt
```diff
@@ -20,12 +20,26 @@ fun Context.getPublicFilesDir() : File = getExternalFilesDir(null) ?: filesDir
...
+    init {
+        instance = this
+    }
+
+    companion object {
+        private var instance: SkylineApplication? = null;
+
+        fun applicationContext() : Context {
+            return instance!!.applicationContext
+        }
```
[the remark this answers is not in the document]
Vixeliz: So I had a little bit of trouble with this since i'm used to static in java however this seems to work:

```
    init {
        instance = this
    }

    companion object {
        lateinit var instance: SkylineApplication
            private set
    }
    
     override fun onCreate() {
        super.onCreate()
        instance = this
     }
    
```

I'm not sure if this is the right way to do so though. Trying to make the context itself a getter led to problems since application context was from the actual class instead of the companion object. This is probably be just being nooby if there is any better way let me know.

app/src/main/java/emu/skyline/provider/DocumentsProvider.kt
```diff
@@ -0,0 +1,269 @@
...
+import android.provider.DocumentsProvider
+import android.webkit.MimeTypeMap
+import emu.skyline.R
+import emu.skyline.SkylineApplication
+import emu.skyline.getPublicFilesDir
+import java.io.*
+
+class DocumentsProvider : DocumentsProvider() {
+
+    private val baseDirectory = File(SkylineApplication.instance.applicationContext.getPublicFilesDir().canonicalPath)
```
AbandonedCart, line 22: In case you didn't already know, `instance` is equivalent to `applicationContext` (much like you can use an activity as its context object, this calls application.applicationContext)

Vixeliz: Oh my bad I didn't know that i'll change that right now with my other fixes

app/src/main/java/emu/skyline/SkylineApplication.kt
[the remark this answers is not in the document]
PixelyIon: I assume this is effectively fixed given the comment below by @AbandonedCart. I did suspect this was the case but wasn't sure as I rarely deal with the application context and almost always have access to the `Activity`.

PixelyIon (changes requested):
A few more nits, just cleaning up the rest of the code.

PR #1423, opened 2022-05-30
Add a lower resolution option 

Joemama45:
If you guys at the skyline team could make this possible, possibly more games could be compatible. I have a powerful device, but a lower resolution option would be very nice to have in case of frequent lag and crashes, so please add this option if it's possible 

Good luck to the skyline team!


HalogenGam3r:
post this on the Issues channel, not the pull requests.

Mr-Hopeless:
pls aad low rsluson mode 🥺🥺🥺

PR #1436, opened 2022-06-01
fix authority name for getUriForFile

dimaxdqwerty:
Authority name should be same as in provider in AndroidManifest.xml. So it fixes crash when trying to get logs


ceelsoin:
yes when i clicked on log i got app crashed

PixelyIon:
I was already working on this locally alongside some other fixes so this PR isn't necessary.

PR #1441, opened 2022-06-01
Make Applet accesses to the data queues lock.

MCredstoner2004:
To avoid potential races when the guest access the same applet from more than one thread


bylaws (approved):

PixelyIon (approved):
LGTM! Sorry for the broken CI, will fix later.

PR #1481, opened 2022-06-05
Stub nfp IUser::AttachAvailabilityChangeEvent .1

Youtube0304:


PixelyIon:
Please refrain from taking commits from `ftx1` and trying to merge them into `master`, none of these commits maintain their original authorship which in this case is @bylaws's commit. If this is continued then we'll have to block you from the repository.

PR #1491, opened 2022-06-07
2D Textures correct

Test0102tr:
After you map the texture with texture->Map(), you try to copy it across into m_pBits all in one go. This won't work unless the RowPitch of the mapped texture is the same as its width


bylaws:
wtf are you??? some AI???

Sovietfowl:
Test Ai.

raiyuulmao:
AI-TEST-BUFFER

PR #1645, opened 2022-06-30
Fix skewed 2D images

ArbestRi:
On some games, like Pokémon Let's Go Eevee (demo version), images are skewed. This plagued many other titles as well. This will fix the image skewing problem.

### DIFF
--- a/app/src/main/cpp/skyline/gpu/texture/texture.cpp
+++ b/app/src/main/cpp/skyline/gpu/texture/texture.cpp
@@ -147,6 +147,14 @@ namespace skyline::gpu {
         if (GetBacking())
             SynchronizeHost();
     }
+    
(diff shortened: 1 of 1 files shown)


PixelyIon:
This will not in-fact fix any skewed textures, it seems like this is generic pitch-linear copy code but copy-pasted into Skyline with no understanding of anything surrounding the code. This isn't even valid C++ as it's not in a function nor do any of the variables that the contents refer to exist. I'm closing this PR as it addresses effectively nothing, it's just entirely invalid code.

headshot2017:
https://stackoverflow.com/questions/15324627/directx-writing-backbuffer-data-to-file my man stole the code from stackoverflow even

ArbestRi:
a

PR #1651, opened 2022-07-01
Write `ApplicationErrorArg` `ErrorApplet`s to log

MCredstoner2004:
These applets are used by applications to display a custom error message to the user. Both the error message and the detailed error message are printed to the error log. (thx @lynxnb)

### DIFF
--- a/app/src/main/cpp/skyline/applet/controller_applet.cpp
+++ b/app/src/main/cpp/skyline/applet/controller_applet.cpp
@@ -15,7 +15,7 @@ namespace skyline::applet {
                                        std::shared_ptr<kernel::type::KEvent> onNormalDataPushFromApplet,
                                        std::shared_ptr<kernel::type::KEvent> onInteractiveDataPushFromApplet,
                                        service::applet::LibraryAppletMode appletMode)
-        : IApplet(state, manager, std::move(onAppletStateChanged), std::move(onNormalDataPushFromApplet), std::move(onInteractiveDataPushFromApplet), appletMode) {}
+        : IApplet{state, manager, std::move(onAppletStateChanged), std::move(onNormalDataPushFromApplet), std::move(onInteractiveDataPushFromApplet), appletMode} {}
 
     void ControllerApplet::HandleShowControllerSupport(input::NpadStyleSet styleSet, ControllerAppletVersion version, span<u8> arg) {
         // Generic macro due to both versions of arguments sharing the same fields but having different layouts
@@ -82,7 +82,7 @@ namespace skyline::applet {
             }
         }
 
-        std::scoped_lock lock{inputDataMutex};
+        std::scoped_lock lock{normalInputDataMutex};
         switch (argPrivate.mode) {
             case ControllerSupportMode::ShowControllerSupport:
                 HandleShowControllerSupport(argPrivate.styleSet, appletVersion, normalInputData.front()->GetSpan());
@@ -100,15 +100,14 @@ namespace skyline::applet {
         // Notify the guest that we've finished running
         onAppletStateChanged->Signal();
         return {};
-    };
+    }
 
     Result ControllerApplet::GetResult() {
         return {};
     }
 
     void ControllerApplet::PushNormalDataToApplet(std::shared_ptr<service::am::IStorage> data) {
-        std::scoped_lock lock{inputDataMutex};
-        normalInputData.emplace(std::move(data));
+        PushNormalInput(data);
     }
 
     void ControllerApplet::PushInteractiveDataToApplet(std::shared_ptr<service::am::IStorage> data) {}
--- a/app/src/main/cpp/skyline/applet/error_applet.cpp
+++ b/app/src/main/cpp/skyline/applet/error_applet.cpp
@@ -1,0 +1,55 @@
+// SPDX-License-Identifier: MPL-2.0
+// Copyright © 2022 Skyline Team and Contributors (https://github.com/skyline-emu/)
+
+#include "error_applet.h"
+#include "services/am/storage/ObjIStorage.h"
+
+namespace skyline::applet {
+    ErrorApplet::ErrorApplet(const DeviceState &state,
+                             service::ServiceManager &manager,
+                             std::shared_ptr<kernel::type::KEvent> onAppletStateChanged,
+                             std::shared_ptr<kernel::type::KEvent> onNormalDataPushFromApplet,
+                             std::shared_ptr<kernel::type::KEvent> onInteractiveDataPushFromApplet,
+                             service::applet::LibraryAppletMode appletMode)
+        : IApplet{state, manager, std::move(onAppletStateChanged), std::move(onNormalDataPushFromApplet), std::move(onInteractiveDataPushFromApplet), appletMode} {}
+
+    Result ErrorApplet::Start() {
+        auto commonArg{PopNormalInput<service::applet::CommonArguments>()};
+
+        errorStorage = PopNormalInput();
+        auto errorCommonHeader{errorStorage->GetSpan().as<ErrorCommonHeader>()};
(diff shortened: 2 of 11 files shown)


app/src/main/cpp/skyline/applet/controller_applet.cpp
```diff
@@ -33,7 +33,8 @@ namespace skyline::applet {
             auto &npad{state.input->npad};
             std::scoped_lock lock{npad.mutex};
 
-            PushNormalDataAndSignal(std::make_shared<service::am::ObjIStorage<ControllerSupportResultInfo>>(state, manager, ControllerSupportResultInfo{
+            PushNormalDataAndSignal(std::make_shared<service::am::ObjIStorage < ControllerSupportResultInfo>>
```
PixelyIon, line 36:
```suggestion
            PushNormalDataAndSignal(std::make_shared<service::am::ObjIStorage<ControllerSupportResultInfo>>
```

app/src/main/cpp/skyline/applet/controller_applet.cpp
```diff
@@ -93,22 +94,22 @@ namespace skyline::applet {
                 normalInputData.pop();
 
                 // Return empty result
-                PushNormalDataAndSignal(std::make_shared<service::am::ObjIStorage<Result>>(state, manager, Result{}));
+                PushNormalDataAndSignal(std::make_shared<service::am::ObjIStorage < Result>>
```
PixelyIon, line 97:
```suggestion
                PushNormalDataAndSignal(std::make_shared<service::am::ObjIStorage<Result>>
```

app/src/main/cpp/skyline/applet/error_applet.h
```diff
@@ -0,0 +1,67 @@
...
+// Copyright © 2020 Ryujinx Team and Contributors (https://github.com/ryujinx/)
+
+#pragma once
+
+#include <services/am/applet/IApplet.h>
+#include "common/language.h"
+
+namespace skyline::applet {
+    /**
+     * @brief The ErrorApplet class is used to show errors to the user
```
PixelyIon, line 12: `The Error Applet is utilized by the guest to display an error to the user, this class reimplements that and prints the error to the log`

app/src/main/cpp/skyline/applet/error_applet.h
```diff
@@ -0,0 +1,67 @@
...
+      private:
+        enum class ErrorType : u8 {
+            ErrorCommonArg,
+            SystemErrorArg,
+            ApplicationErrorArg,
+            ErrorEulaArg,
+            ErrorPctlArg,
+            ErrorRecordArg,
+            SystemUpdateEulaArg = 8
+        };
```
PixelyIon, line 26: I'd recommend setting every member of `ErrorType`, this is the general practice with guest enums.

app/src/main/cpp/skyline/applet/error_applet.h
```diff
@@ -0,0 +1,67 @@
...
+            ErrorType type;
+            u8 jump;
+            u8 _pad_[0x3];
+            u8 contextFlag;
+            u8 resultFlag;
+            u8 contextFlag2;
+        };
+        static_assert(sizeof(ErrorCommonHeader) == 0x8);
+
+        static_assert(sizeof(u64) == 8);
```
PixelyIon, line 41: There is no need for this, you can assume it's true. I'm guessing this was left over from debugging.

app/src/main/cpp/skyline/applet/error_applet.h
```diff
@@ -0,0 +1,67 @@
...
+            ErrorEulaArg,
+            ErrorPctlArg,
+            ErrorRecordArg,
+            SystemUpdateEulaArg = 8
+        };
+
+        /**
+         * @url https://switchbrew.org/wiki/Error_Applet#ErrorCommonHeader
+         */
+        struct __attribute__((packed)) ErrorCommonHeader {
```
PixelyIon, line 31: Use the `#pragma` rather than the `__attribute__` here since it covers multiple consecutive structs.

app/src/main/cpp/skyline/applet/error_applet.cpp
```diff
@@ -0,0 +1,52 @@
...
+        // Notify the guest that we've finished running
+        onAppletStateChanged->Signal();
+
+        return {};
+    }
+
+    void ErrorApplet::HandleApplicationErrorArg() {
+        auto applicationErrorStorage{errorStorage->GetSpan().as<ApplicationErrorArg>()};
+
+        Logger::Error("{}\nFull message: {}", applicationErrorStorage.dialogMessage.data(), applicationErrorStorage.fullscreenMessage.data());
```
PixelyIon, line 40: Isn't it possible for `Full Message` to be unset? It shouldn't really be printed in that case. I'd also recommend `ErrorNoPrefix("Application Error: ...")` as it'd be better than `HandleApplicationErrorArg`.

app/src/main/cpp/skyline/services/am/applet/IApplet.h
```diff
@@ -70,4 +69,60 @@ namespace skyline::service::am {
          */
         std::shared_ptr<IStorage> PopInteractiveAndClear();
     };
+
+    /**
+     * @brief Utility class for applets that need to queue the normal data sent to them
+     */
+    class EnableNormalDataQueue {
```
PixelyIon, line 76: This should be renamed to `NormalQueueApplet`.

app/src/main/cpp/skyline/services/am/applet/IApplet.h
```diff
@@ -70,4 +69,60 @@ namespace skyline::service::am {
...
+
+        void PushNormalInput(std::shared_ptr<service::am::IStorage> data) {
+            normalInputData.emplace(data);
+        }
+    };
+
+    /**
+     * @brief Utility class for applets that need to queue the interactive data sent to them
+     */
+    class EnableInteractiveDataQueue {
```
PixelyIon, line 104: This should be renamed to `InteractiveQueueApplet`.

PixelyIon (changes requested):
A few nits here and there.

app/src/main/cpp/skyline/applet/error_applet.cpp
lynxnb: As a stub, I'd say it's not too bad to print `Full message:` even when one isn't supplied. Conditionally appending that would be annoying IMO as you'd have to format locally, then submit to logger. We can do this properly later on.

PixelyIon: That's not required, it can simply be:
```suggestion
    if (applicationErrorStorage.fullscreenMessage.empty())
        Logger::ErrorNoPrefix("Application Error: {}", applicationErrorStorage.dialogMessage.data());
    else
        Logger::ErrorNoPrefix("Application Error: {}\nFull message: {}", applicationErrorStorage.dialogMessage.data(), applicationErrorStorage.fullscreenMessage.data());
```

PixelyIon: That's not required, it can simply be:
```suggestion
    if (applicationErrorStorage.fullscreenMessage.empty())
        Logger::ErrorNoPrefix("Application Error: {}", applicationErrorStorage.dialogMessage.data());
    else
        Logger::ErrorNoPrefix("Application Error: {}\nFull message: {}", applicationErrorStorage.dialogMessage.data(), applicationErrorStorage.fullscreenMessage.data());
```

lynxnb: Well I didn't think of that for some reason, lgtm!

PixelyIon (approved):
LGTM!

PR #1663, opened 2022-07-04
Add Chinese text

neighbor-bear:
Added some Chinese text, hoping to add it to the simulator

PR #1672, opened 2022-07-06
Translation pt_BR strings.xml

HERCULESxp:
Translation of strings to Brazilian Portugueses [pt_BR]


PixelyIon:
See #1663, these changes are useful. It's just that we'll have a better way to submit them in the future.

PR #1681, opened 2022-07-08
Update BUILDING.md

HERCULESxp:
ndk path windows  to add into local.propertes

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -109,6 +109,10 @@ Put path to the NDK in `ndk.dir` property of `local.properties` file, eg.:
 ```properties
 ndk.dir=/home/skyline/opt/android-sdk/ndk/25.0.8221429
 ```
+On Windows:
+```properties
+ndk.dir=C\:\\Users\\skyline\\AppData\\Local\\Android\\Sdk\\ndk\\25.0.8221429
+```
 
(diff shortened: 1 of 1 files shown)


mjcox244:
Unfortunately this update to the guide will not work for anyone with a user account not called "skyline"

finr0y:
if you change it to %USERPROFILE% i think it can work for anyone

HERCULESxp:
I put it in the same way as above with Linux, but using the %USERPROFILE% It's the best way.

mjcox244:
> I put it in the same way as above with Linux, but using the %USERPROFILE% It's the best way.

that's a point actually, why isn't it using "$HOME"

HERCULESxp:
what would the path be like in linux? It would be like this: ```properties
ndk.dir=/$HOME/opt/android-sdk/ndk/25.0.8221429
```

bylaws:
ndk.dir is deprecated and should not be used

AbandonedCart:
https://github.com/skyline-emu/skyline/commit/adba8c52c067d0c495e5f796b41071ea2ac0fb10 is a bit misleading.

Linux should be $HOME, while Mac is ~, and Windows is %USERPROFILE% but it doesn't specify a platform and follows an example given for Windows.

PR #1735, opened 2022-07-31
Add RGB565 format workaround

guranu:
Hi, sorry to be using pull request incorrectly but is it possible to add back RGB565 workaround? Again i am very sorry for using pull requests incorrectly.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -25,7 +25,7 @@ git clone https://github.com/skyline-emu/skyline.git --recursive
 ```
 
 Open Android Studio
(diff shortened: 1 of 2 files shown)


PixelyIon:
I'm sorry but knowing something shouldn't be done but doing it anyway isn't an excuse for doing it, our Discord would've been the more appropriate forum for this. I'm unsure what you're referring to as the RGB565 workaround, that could mean many things but it's very likely that it's not going to be added back.

guranu:
How do you not know about this:
Screenshot_20220801-094723_1

PR #1894, opened 2022-10-08
Android 8 Compatibility

Kawata0210:
Add android 8 compatibility for Older device can use Skyline™


PixelyIon:
We will never be adding Android 8 support, most devices that are on Android 8 will not be able to run Skyline regardless.

SpyroGam3r:
Buy a phone with Android 10 eta when :trollface:

mateoalfaro:
> Buy a phone with Android 10 eta when :trollface:

Accept android 9 superiority when smh

nekomekoraiyuu:
Bro just use a damn gsi
(Instant mali profit) :jefftrollo:

Levella2000:
does not work on android 9

PR #1922, opened 2022-10-25
Run emulation in a separate process

lynxnb:


AbandonedCart:
Working as intended so far when tested against games that previously terminated the entire app.

PR #1937, opened 2022-10-30
Store KMemory object ptrs in memory class to avoid linear-time unmap

bylaws:
This is quite a horrible solution but fixing it properly would require a whole rewrite of how we handle memory. Vastly reduces stutters in skyrim from searching over thousands of handles every unmap.


AbandonedCart:
Adding `#include <common/base.h>` to KProcess.cpp will fix the error, although I am still not entirely sure why it didn't inherit the definition already.

PR #1941, opened 2022-10-31
navbar sticking out workaround for certain android roms

lybxlpsv:
Some older android roms may have broken immersive mode implementation that causes navigation bar to stick out indefinitely after touching the screen.
This workaround should fix it.

### DIFF
--- a/app/src/main/java/emu/skyline/EmulationActivity.kt
+++ b/app/src/main/java/emu/skyline/EmulationActivity.kt
@@ -344,7 +344,7 @@ class EmulationActivity : AppCompatActivity(), SurfaceHolder.Callback, View.OnTo
 
         changeAudioStatus(true)
 
-        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.R) {
+        if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.R) {
             @Suppress("DEPRECATION")
             window.decorView.systemUiVisibility = (View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY
                     or View.SYSTEM_UI_FLAG_LAYOUT_STABLE
--- a/app/src/main/java/emu/skyline/input/onscreen/OnScreenEditActivity.kt
+++ b/app/src/main/java/emu/skyline/input/onscreen/OnScreenEditActivity.kt
@@ -121,7 +121,7 @@ class OnScreenEditActivity : AppCompatActivity() {
     }
 
     private fun fullScreen() {
-        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.R) {
+        if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.R) {
             @Suppress("DEPRECATION")
             window.decorView.systemUiVisibility = (View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY
                     or View.SYSTEM_UI_FLAG_LAYOUT_STABLE


lynxnb:
> Some older android roms
That is actually incorrect, it's some _newer_ Android ROMs that have the broken behaviour. By trying to fix the issue for a few devices, we're basically breaking spec and using deprecated code on working ones. Honestly, I'd rather have a setting that enables the old way of hiding system bars.

bylaws:
> > Some older android roms
> 
> That is actually incorrect, it's some _newer_ Android ROMs that have the broken behaviour. By trying to fix the issue for a few devices, we're basically breaking spec and using deprecated code on working ones. Honestly, I'd rather have a setting that enables the old way of hiding system bars.

I'd rather use this as it doesn't require any settings and works fine

lybxlpsv:
if deprecation is an issue. it is also possible to use

`window.insetsController?.let {
                it.systemBarsBehavior = WindowInsetsController.BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE
 }`

for a11, and the deprecated window.decorView.systemUiVisibility for a10.

AbandonedCart:
This is a mess because Android has changed the API so many times to accomplish the same thing. Each time, they neglected to leave an effective backwards compatibility. If you need a solution for API < 11, why not a listener?

lybxlpsv:
> If you need a solution for API < 30,

btw we need solution for API <= 30, not just API < 30.
Anyways I tried your snippet on A11 with adjusted SDK INT check, tested on both before and after setcontentview, and navbar still sticks on my device.

AbandonedCart:
> btw we need solution for API <= 30, not just API < 30.

R (30) no longer uses those methods, it uses the ones already implemented.

lybxlpsv:
> > btw we need solution for API <= 30, not just API < 30.
> 
> R (30) no longer uses those methods, it uses the ones already implemented.
> 
> #1942 <- Easier to demonstrate than explain

The main point of this PR is to workaround the navbar sticking issue.
Your PR is removal of deprecated function into compat.
It does not fix the navbar sticking issues, meaning it is not an alternative to this PR.

But it does clean things up for the workaround I'm proposing.

```
override fun onWindowFocusChanged(hasFocus : Boolean) {
	super.onWindowFocusChanged(hasFocus)
	if (hasFocus) \\ maybe limit workaround to R and below only?
	{
		WindowInsetsControllerCompat(window, window.decorView).let { controller ->
			controller.systemBarsBehavior = WindowInsetsControllerCompat.BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE
		}
	}
}
```

lynxnb:
As it currently stands, Skyline already respects the deprecation notices and best practices from the Android documentation regarding hiding the system bars. The only change I'd say we might make is using the view flags method up until Android 11 instead of 10.

In practice, drop all commits and just make one that replaces `<` with `<=` in the onResume android sdk level check. We might as well drop #1942.

AbandonedCart:
@lynxnb Actually, the documentation contradicts you. https://developer.android.com/develop/ui/views/layout/immersive#kotlin

lynxnb:
In what way does it contradict what I said? We don't make use of the Compat classes but instead make direct calls to WindowInsetController and handle legacy flags ourselves, which is functionally the same.

AbandonedCart:
> In what way does it contradict what I said? We don't make use of the Compat classes but instead make direct calls to WindowInsetController and handle legacy flags ourselves, which is functionally the same.

> Skyline already respects the deprecation notices and **best practices from the Android documentation** regarding hiding the system bars

It's your choice (and that's fine), but it's not the best practices of the documentation. It's only fair to separate the two.

GuilhermeGS2:
Thank you very much, that fixed the issue I had here with ArrowOS v11.0 on my Poco F3. Seems to be working perfectly fine now.

lynxnb:
Thanks for the report!

lynxnb:
PR is good to go after you fix here: https://github.com/skyline-emu/skyline/blob/ftx1/app/src/main/java/emu/skyline/input/onscreen/OnScreenEditActivity.kt#L124

lynxnb (approved):

PR #1943, opened 2022-10-31
Stub nsd:u/nsd:a and sfdnsres services. Stub some bsd functions

dimaxdqwerty:
Required by some games to boot (e.g. No Man's Sky)


app/src/main/cpp/skyline/services/socket/bsd/IClient.cpp
```diff
@@ -15,7 +15,70 @@ namespace skyline::service::socket {
         return {};
     }
 
+    Result IClient::Select(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
```
bylaws, line 18: Commit message for this is broken, needs a newline after title

PR #1960, opened 2022-11-05
Miscellaneous UI fixes

lynxnb:
Fixes small design inconsistencies and layout bugs across the app.

See individual commits description for more.


app/src/main/java/emu/skyline/AppDialog.kt
```diff
@@ -101,5 +82,14 @@ class AppDialog : BottomSheetDialogFragment() {
 
             shortcutManager.requestPinShortcut(info.build(), null)
         }
+
+        dialog?.setOnKeyListener { _, keyCode, event ->
+            if (keyCode == KeyEvent.KEYCODE_BUTTON_B && event.action == KeyEvent.ACTION_UP) {
+                dialog?.onBackPressed()
+                true
+            } else {
```
dimaxdqwerty, line 90: Is this `else` really needed? It will be `false` anyway without it

lynxnb: It's needed because of how returning from lambda works in Kotlin.

PixelyIon (approved):
LGTM!

PR #1971, opened 2022-11-09
Improve robustness of `KeyReader.import`

lynxnb:
* Close the input and output file streams before moving the output file to the final destination
* Clean up the destination path before moving the new file
* Introduce a `ImportResult` return value to differentiate between the possible causes of import errors
* Display more meaningful error messages in the UI


bylaws:
line endings buggered

PixelyIon (approved):

PR #1999, opened 2022-11-17
Work around invalid values in `CNTFRQ_EL0` register

lynxnb:
Exynos SoCs have a bug where the `CNTFRQ_EL0` register is either set to 0 or contain incoherent values. With this patch, the frequency value is loaded into a static variable and used instead of reading the register. The value will be initialised to the correct value for affected SoCs, while unaffected ones will use the value from the register.

### DIFF
--- a/app/src/main/cpp/skyline/common/utils.h
+++ b/app/src/main/cpp/skyline/common/utils.h
@@ -8,6 +8,7 @@
 #include <span>
 #include <frozen/unordered_map.h>
 #include <frozen/string.h>
+#include <sys/system_properties.h>
 #include <type_traits>
 #include <xxhash.h>
 #include "base.h"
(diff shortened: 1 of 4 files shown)


app/src/main/cpp/skyline/common/utils.h
```diff
@@ -20,13 +21,38 @@ namespace skyline::util {
...
+            else if (board == "exynos9810")     // Exynos 9810
+                frequency = 26000000;
+            else
+                asm volatile("MRS %0, CNTFRQ_EL0" : "=r"(frequency));
+
+            return frequency;
+        }
+    }
+
+    inline const u64 clockFrequency{detail::InitFrequency()}; //!< The system counter clock frequency in Hz
```
PixelyIon, line 48: Globals should be `CamelCase`, fix all usages of this too.
```suggestion
    inline const u64 ClockFrequency{detail::InitFrequency()}; //!< The system counter clock frequency in Hz
```

app/src/main/cpp/skyline/nce/guest.cpp
```diff
@@ -0,0 +1,45 @@
...
+        *code++ = 0x8B010421; // ADD X1, X1, X1, LSL #1
+        *code++ = 0xD37AE420; // LSL X0, X1, #6
+
+        /* Store result */
+        /* Restore registers */
+        *code++ = 0xF90003E0; // STR X0, [SP, #0]
+        *code++ = 0xA94107E0; // LDP X0, X1, [SP, #16]
+
+        return code;
+    }
```
PixelyIon, line 44: No need for this, just put it in `nce.cpp` as we do for all codegen.

app/src/main/cpp/skyline/nce/guest.h
```diff
@@ -105,7 +105,7 @@ namespace skyline {
         namespace guest {
             constexpr size_t SaveCtxSize{34}; //!< The size of the SaveCtx function in 32-bit ARMv8 instructions
             constexpr size_t LoadCtxSize{34}; //!< The size of the LoadCtx function in 32-bit ARMv8 instructions
-            constexpr size_t RescaleClockSize{16}; //!< The size of the RescaleClock function in 32-bit ARMv8 instructions
+            constexpr size_t RescaleClockSize{17}; //!< The size of the RescaleClock function in 32-bit ARMv8 instructions
```
PixelyIon, line 108: Move this to `nce.cpp` as well.

PixelyIon (changes requested):

PixelyIon (approved):
LGTM! If someone can test then this is good to go.

PR #2004, opened 2022-11-19
Implement compute shaders

bylaws:
This can reuse a fair bit of the Maxwell 3D code which has now been commonised. Of note is how compute lacks any active state at all due to its use of QMDs which bundle up all state into a single object in memory. Aside from that see the commit messages for more info.


app/src/main/cpp/skyline/soc/gm20b/gpfifo.cpp
```diff
@@ -359,9 +359,11 @@ namespace skyline::soc::gm20b {
...
                 // If we run out of GpEntries to process ensure we submit any remaining GPU work before waiting for more to arrive
                 Logger::Debug("Finished processing pushbuffer batch");
-                channelCtx.executor.Submit();
-                channelCtx.Unlock();
-                channelLocked = false;
+                if (channelLocked) {
+                    channelCtx.executor.Submit();
+                    channelCtx.Unlock();
+                    channelLocked = false;
+                }|
```
AbandonedCart, line 366: I believe the pipe here may be a typo.

Issue #2009, opened 2022-11-21
Haptic feedback - On-screen buttons - Vibration


Dougzen:
Hello guys.

As recommended, i'm making a new issue:

Would be interesting to have the Haptic Feedback for On-Screen buttons. Vibration feature.

Example:
If the "A" button be pressed On-Screen, you could feel the device vibrating.

Thank you for your attention.

Dougzen:
Okay I guess, someone had recommended me to make a "issue" instead a PR.

AbandonedCart:
@Dougzen Yes, you made the issue and I submitted a PR to add the feature.

Swizzlefowl:
Merged on the latest Build.

Issue #1989, opened 2022-11-15
need to add the L3 and R3 to the on-screen controls


pokeholmes001:
need to add the L3 and R3 to the on-screen controls

bylaws:
double tap sticks

PR #2010, opened 2022-11-21
Add a rudimentary OSC button vibration setting

AbandonedCart:
No frills vibration for OSC buttons. Toggled from input settings, only activates for OSC, doesn't apply to joysticks. Vibration length / intensity mirrors pressing one of the navigation buttons (home, back, etc). Fulfills #2009

### DIFF
--- a/app/src/main/java/emu/skyline/input/onscreen/OnScreenControllerView.kt
+++ b/app/src/main/java/emu/skyline/input/onscreen/OnScreenControllerView.kt
@@ -66,6 +66,13 @@ class OnScreenControllerView @JvmOverloads constructor(context : Context, attrs 
             field = value
             (controls.circularButtons + controls.rectangularButtons + controls.triggerButtons).forEach { it.hapticFeedback = hapticFeedback }
         }
+    private val vibrator: Vibrator =
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+            (context.getSystemService(VIBRATOR_MANAGER_SERVICE) as VibratorManager).defaultVibrator
+        } else {
+            @Suppress("DEPRECATION") (context.getSystemService(VIBRATOR_SERVICE) as Vibrator)
+        }
+    private val effectClick = VibrationEffect.createPredefined(VibrationEffect.EFFECT_CLICK)
 
     override fun onDraw(canvas : Canvas) {
         super.onDraw(canvas)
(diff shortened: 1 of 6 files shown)


app/src/main/java/emu/skyline/input/onscreen/OnScreenControllerView.kt
```diff
@@ -286,4 +298,15 @@ class OnScreenControllerView @JvmOverloads constructor(context : Context, attrs
         controls.allButtons.first { it.buttonId == buttonId }.config.enabled = enabled
         invalidate()
     }
+
+    private fun vibrate() {
+        val vibe = VibrationEffect.createOneShot(20, VibrationEffect.DEFAULT_AMPLITUDE)
```
lynxnb, line 303: Probably best to save this as member instead of recreating it every time.

lynxnb (commented):
Looking good and clean. I'm not sure if we really want haptic feedback to be ON by default, I'd rather have it OFF.

AbandonedCart:
> Looking good and clean. I'm not sure if we really want haptic feedback to be ON by default, I'd rather have it OFF.

Meanwhile, I've never understood how it hasn't become an attribute of buttons in the Android APIs yet.

AbandonedCart:
There isn't a whole mess of evidence that caching the entire function to a variable is any more or less efficient than running it as a function, but if there is any chance... might as well. It's not like anything actually changes between calls.

bylaws:
Would be nice if sticks had it tbh, just for the initial press if that's doable

AbandonedCart:
@bylaws I would definitely recommend not attaching vibration to the sticks on a system that has an R3 / L3 control scheme. The vibration is misleading because it implies the push in action over movement.

bylaws:
Ez

> @bylaws I would definitely recommend not attaching vibration to the sticks on a system that has an R3 / L3 control scheme. The vibration is misleading because it implies the push in action 

Speaking of which, do you fancy making seperate r/l3 buttons since some games require them? I get your point though so fair enough rn

dima-xd:
How do L3/R3 buttons will even look in OSC? Like optional buttons at bottom?

Swizzlefowl:
Reopened #1989.

mboforward1:
Le mar. 22 nov. 2022 à 14:58, Swizzlefowl ***@***.***> a
écrit :

lynxnb (approved):

bylaws:
can you make the vibrations slightly lighter? I think it would help

AbandonedCart:
It's actually not set. The current intensity is what you have configured as the system vibration intensity in the Settings app. These are the other options -

int | EFFECT_CLICK           A click effect.
int | EFFECT_DOUBLE_CLICK           A double click effect.
int | EFFECT_HEAVY_CLICK           A heavy click effect.
int | EFFECT_TICK           A tick effect.

Keep in mind that setting any one of these will NOT reflect the setting for the device.

AbandonedCart:
I changed it to click to see if that is any better. Unfortunately, the word "click" doesn't give a very good preview of how it will actually feel without firing it up.

AbandonedCart:
I didn't test `EFFECT_TICK` after how subtle `EFFECT_CLICK` already was, but `EFFECT_HEAVY_CLICK` didn't produce any vibration on either of my devices and `EFFECT_DOUBLE_CLICK` is two separate (albeit barely) vibrations. `EFFECT_CLICK` is probably the only option.

PixelyIon:
@bylaws Can you retest so this is RTM?

najshesh:
How can i turn off the vibration??

AbandonedCart:
Same way you turned it on in settings.

najshesh:
> Same way you turned it on in settings.

I did't turned it on 😐
When i installed latest version 
It automatically turned on.

AbandonedCart:
This isn't merged, so it couldn't be from this.

PR #2014, opened 2022-11-22
Add TitleId TextView in App Dialog

dima-xd:
Also if you hold with touch the Title ID it will copy it to clipboard

### DIFF
--- a/app/src/main/java/emu/skyline/AppDialog.kt
+++ b/app/src/main/java/emu/skyline/AppDialog.kt
@@ -19,6 +19,7 @@ import androidx.core.content.ContextCompat
 import androidx.core.graphics.drawable.toBitmap
 import com.google.android.material.bottomsheet.BottomSheetBehavior
 import com.google.android.material.bottomsheet.BottomSheetDialogFragment
+import com.google.android.material.snackbar.Snackbar
 import emu.skyline.data.AppItem
 import emu.skyline.databinding.AppDialogBinding
 import emu.skyline.loader.LoaderResult
@@ -66,6 +67,7 @@ class AppDialog : BottomSheetDialogFragment() {
         binding.gameIcon.setImageBitmap(item.icon ?: missingIcon)
         binding.gameTitle.text = item.title
         binding.gameVersion.text = item.version ?: item.loaderResultString(requireContext())
+        binding.gameTitleId.text = item.titleId
         binding.gameAuthor.text = item.author
 
         binding.gamePlay.isEnabled = item.loaderResult == LoaderResult.Success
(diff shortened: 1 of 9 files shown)


AbandonedCart:
Out of curiosity, what is the purpose of having it copy to the clipboard?

dima-xd:
> 

For example when you need to know Title ID of the game to import save file

AbandonedCart:
Oh ok. That makes sense. It is a pain to look them up.

AndresNorm:
Well, this is great.
Less hassle for googling the ID now

app/src/main/java/emu/skyline/AppDialog.kt
```diff
@@ -91,6 +92,12 @@ class AppDialog : BottomSheetDialogFragment() {
             shortcutManager.requestPinShortcut(info.build(), null)
         }
 
+        binding.gameTitleId.setOnLongClickListener {
+            val clipboard = requireActivity().getSystemService(android.content.Context.CLIPBOARD_SERVICE) as android.content.ClipboardManager
+            clipboard.setPrimaryClip(android.content.ClipData.newPlainText("Title ID", item.titleId))
```
lynxnb, line 97: Show a snack bar after you do this saying "Copied to clipboard".

lynxnb (changes requested):

app/src/main/java/emu/skyline/AppDialog.kt
dima-xd: > Show a snack bar after you do this saying "Copied to clipboard".

Done. Also added this line to `strings.xml`

app/src/main/res/values/strings.xml
```diff
@@ -194,6 +194,7 @@
     <!-- Software Keyboard -->
     <string name="input_hint">Input Text</string>
     <!-- Misc -->
+    <string name="copied_to_clipboard">Copied to clipboard</string>
```
lynxnb, line 197: Don't make a new Misc section for this, but move this at the top of the file, in the Common section.

lynxnb (changes requested):

app/src/main/res/values/strings.xml
dima-xd: > Don't make a new Misc section for this, but move this at the top of the file, in the Common section.

It's not new btw, it was just an empty section

lynxnb: Right, in that case delete it, and move to Common.

dima-xd: Done

lynxnb (approved):

PR #2054, opened 2022-12-05
Add support for selecting a profile picture

PabloG02:
I took as a base dima-xd's stub to implement it.
Results:

Screenshot_20221205_192849


app/src/main/cpp/skyline/services/account/IProfile.cpp
```diff
@@ -3,8 +3,22 @@
...
+        0xFF, 0xD8, 0xFF, 0xDB, 0x00, 0x43, 0x00, 0x03, 0x02, 0x02, 0x02, 0x02,
+        0x02, 0x03, 0x02, 0x02, 0x02, 0x03, 0x03, 0x03, 0x03, 0x04, 0x06, 0x04,
+        0x04, 0x04, 0x04, 0x04, 0x08, 0x06, 0x06, 0x05, 0x06, 0x09, 0x08, 0x0A,
+        0x0A, 0x09, 0x08, 0x09, 0x09, 0x0A, 0x0C, 0x0F, 0x0C, 0x0A, 0x0B, 0x0E,
+        0x0B, 0x09, 0x09, 0x0D, 0x11, 0x0D, 0x0E, 0x0F, 0x10, 0x10, 0x11, 0x10,
+        0x0A, 0x0C, 0x12, 0x13, 0x12, 0x10, 0x13, 0x0F, 0x10, 0x10, 0x10, 0xFF,
+        0xC9, 0x00, 0x0B, 0x08, 0x00, 0x01, 0x00, 0x01, 0x01, 0x01, 0x11, 0x00,
+        0xFF, 0xCC, 0x00, 0x06, 0x00, 0x10, 0x10, 0x05, 0xFF, 0xDA, 0x00, 0x08,
+        0x01, 0x01, 0x00, 0x00, 0x3F, 0x00, 0xD2, 0xCF, 0x20, 0xFF, 0xD9
+    };
```
PixelyIon, line 20: We probably just want to pack a decent default profile picture, I think the Skyline icon might be a decent candidate. It can just be in Android asset FS rather than hardcoded in code.

app/src/main/cpp/skyline/services/account/IProfile.cpp
```diff
@@ -38,4 +52,40 @@ namespace skyline::service::account {
...
     }
+
+    Result IProfile::GetImageSize(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
+        std::vector<char> profileImageIcon = GetProfilePicture();
+        if(profileImageIcon.empty()){
+            response.Push<u32>(profileImageIconDefault.size());
+        } else {
+            response.Push<u32>(profileImageIcon.size());
+        }
+        return {};
```
PixelyIon, line 63: This doesn't match our coding guidelines, I'd suggest giving it a quick read. It's also not a good idea to read a whole file to get its size.
```suggestion
        std::vector<char> profileImageIcon{GetProfilePicture()};
        if (profileImageIcon.empty())
            response.Push<u32>(profileImageIconDefault.size());
        else
            response.Push<u32>(profileImageIcon.size());
        return {};
```

PixelyIon (changes requested):
Great job on the initial draft, I'd say this does need to be restructured a bit in terms of how we get the image and what the default image should be.

app/src/main/cpp/skyline/services/account/IProfile.cpp
```diff
@@ -38,4 +52,40 @@ namespace skyline::service::account {
...
+        if(profileImageIconPointer.is_open()){
+            std::streamsize size = profileImageIconPointer.tellg();
+            profileImageIconPointer.seekg(0, std::ios::beg);
+            std::vector<char> profileImageIcon(size);
+            profileImageIconPointer.read(profileImageIcon.data(), size);
+            profileImageIconPointer.close();
+            return profileImageIcon;
+        }
+        return {};
+    }
```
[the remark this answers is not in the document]
bylaws: I wouldn't bother with using fs here for user selected one as android only grants us permission to the single selected file afaik? And that kinda conflicts with stuff

app/src/main/cpp/skyline/services/account/IProfile.cpp
```diff
@@ -53,14 +44,30 @@ namespace skyline::service::account {
...
     }
 
     Result IProfile::LoadImage(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
-        // TODO: load actual profile image
-        request.outputBuf.at(0).copy_from(profileImageIcon);
-        response.Push<u32>(profileImageIcon.size());
+        std::shared_ptr<vfs::Backing> profileImageIcon{GetProfilePicture()};
+        std::vector<char> profileImage(profileImageIcon->size);
+
+        profileImageIcon->Read(skyline::span<char>(profileImage), 0);
```
bylaws, line 57:
```suggestion
        profileImageIcon->Read(span<char>(profileImage), 0);
```

app/src/main/cpp/skyline/services/account/IProfile.cpp
```diff
@@ -53,14 +44,30 @@ namespace skyline::service::account {
...
+
+        request.outputBuf.at(0).copy_from(profileImage);
+        response.Push<u32>(profileImageIcon->size);
```
bylaws, line 60:
```suggestion
        response.Push(static_cast<u32>(profileImageIcon->size));
```

app/src/main/cpp/skyline/services/account/IProfile.cpp
```diff
@@ -53,14 +44,30 @@ namespace skyline::service::account {
...
+
         return {};
     }
+
+    std::shared_ptr<vfs::Backing> IProfile::GetProfilePicture(){
+        const std::string profilePicturePath{*state.settings->profilePictureValue};
```
bylaws, line 66:
```suggestion
        std::string profilePicturePath{*state.settings->profilePictureValue};
```

bylaws (changes requested):

app/src/main/cpp/skyline/services/account/IProfile.cpp
```diff
@@ -53,14 +44,30 @@ namespace skyline::service::account {
...
     Result IProfile::LoadImage(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
-        // TODO: load actual profile image
-        request.outputBuf.at(0).copy_from(profileImageIcon);
-        response.Push<u32>(profileImageIcon.size());
+        std::shared_ptr<vfs::Backing> profileImageIcon{GetProfilePicture()};
+        std::vector<char> profileImage(profileImageIcon->size);
+
+        profileImageIcon->Read(skyline::span<char>(profileImage), 0);
+
+        request.outputBuf.at(0).copy_from(profileImage);
```
PixelyIon, line 59: Rather than reading it into `profileImage` first then copying it into the output buffer, you can directly read it into `outputBuf`.

app/src/main/cpp/skyline/services/account/IProfile.cpp
```diff
@@ -53,14 +44,30 @@ namespace skyline::service::account {
...
+        response.Push<u32>(profileImageIcon->size);
+
         return {};
     }
+
+    std::shared_ptr<vfs::Backing> IProfile::GetProfilePicture(){
+        const std::string profilePicturePath{*state.settings->profilePictureValue};
+        int fd{open((profilePicturePath).c_str(), O_RDONLY)};
+        if (fd < 0)
+            return state.os->assetFileSystem->OpenFile("skyline-logo.jpeg");
```
PixelyIon, line 69:
```suggestion
        if (fd < 0)
            // If we can't find the profile picture then just return the placeholder profile picture
            return state.os->assetFileSystem->OpenFile("skyline-logo.jpeg");
```

app/src/main/java/emu/skyline/preference/ImagePickerPreference.kt
```diff
@@ -0,0 +1,83 @@
...
+import emu.skyline.SkylineApplication
+import emu.skyline.getPublicFilesDir
+import java.io.File
+import java.io.FileOutputStream
+import java.io.InputStream
+
+class ImagePickerPreference @JvmOverloads constructor(context : Context, attrs : AttributeSet? = null, defStyleAttr : Int = R.attr.preferenceStyle) : Preference(context, attrs, defStyleAttr) {
+    private val pickMedia = (context as ComponentActivity).registerForActivityResult(ActivityResultContracts.PickVisualMedia()) { uri ->
+        val profilePictureDir = SkylineApplication.instance.getPublicFilesDir().canonicalPath + "/switch/nand/system/save/8000000000000010/su/avators"
+        val profilePictureName = "profile_picture.jpeg"
```
PixelyIon, line 31: We'd probably want the path here to be different, not in NAND, as VFS controls that. @lynxnb opinions on where this should be?

app/src/main/java/emu/skyline/preference/ImagePickerPreference.kt
```diff
@@ -0,0 +1,83 @@
...
+import androidx.preference.Preference.SummaryProvider
+import androidx.preference.PreferenceManager
+import androidx.preference.R
+import emu.skyline.SkylineApplication
+import emu.skyline.getPublicFilesDir
+import java.io.File
+import java.io.FileOutputStream
+import java.io.InputStream
+
+class ImagePickerPreference @JvmOverloads constructor(context : Context, attrs : AttributeSet? = null, defStyleAttr : Int = R.attr.preferenceStyle) : Preference(context, attrs, defStyleAttr) {
```
PixelyIon, line 28: This class isn't a generic image picker and it's fair as to why but to reflect this it should be named something on the lines of `ProfilePicturePreference`.

app/src/main/java/emu/skyline/preference/ImagePickerPreference.kt
```diff
@@ -0,0 +1,83 @@
...
+    private val pickMedia = (context as ComponentActivity).registerForActivityResult(ActivityResultContracts.PickVisualMedia()) { uri ->
+        val profilePictureDir = SkylineApplication.instance.getPublicFilesDir().canonicalPath + "/switch/nand/system/save/8000000000000010/su/avators"
+        val profilePictureName = "profile_picture.jpeg"
+        try{
```
PixelyIon, line 32: There are quite a few formatting issues like no space before the bracket here that should be handled by Android Studio's formatter, can you run it on the file?

app/src/main/java/emu/skyline/preference/ImagePickerPreference.kt
```diff
@@ -0,0 +1,83 @@
...
+            if (uri != null) {  // The user selected a photo.
```
PixelyIon, line 33: We don't tend to include full stops at the ends of comments unless it's a part of a larger statement, also for consistency's sake everything should use `picture` rather than `photo`.

app/src/main/java/emu/skyline/preference/ImagePickerPreference.kt
```diff
@@ -0,0 +1,83 @@
...
+                    val readUriPermission: Int = Intent.FLAG_GRANT_READ_URI_PERMISSION
+                    contentResolver.takePersistableUriPermission(uri, readUriPermission)
+                    contentResolver.openInputStream(uri)?.use { inputStream: InputStream ->
+                        var bitmap = BitmapFactory.decodeStream(inputStream)
+                        // Compress the image.
+                        bitmap = Bitmap.createScaledBitmap(bitmap,256,256,false)
+                        StoreBitmap(bitmap, "$profilePictureDir/$profilePictureName")
+                    }
+                }
+            } else {    // User didn't select a photo. As such, if the user already had one, it's assumed that he wants to remove it.
```
PixelyIon, line 46:
```suggestion
            } else { // No picture was selected, clear the profile picture if one was already set
```

app/src/main/java/emu/skyline/preference/ImagePickerPreference.kt
```diff
@@ -0,0 +1,83 @@
...
+            }
+            notifyChanged()
+        } catch (e: Exception){
+            e.printStackTrace()
+        }
+    }
+
+    init {
+        summaryProvider = SummaryProvider<ImagePickerPreference> { preference ->
+            Uri.decode(preference.getPersistedString("No photo selected"))
```
PixelyIon, line 60: `picture`*, you can do so for all instances of it.

PixelyIon (changes requested):
You've done a really solid job at addressing previous criticism, I've caught some with newer code but also a lot with the Kotlin code. I'll add on @lynxnb to properly review the Kotlin side of affairs.

bylaws (approved):

PixelyIon (approved):
Perfectly done, LGTM!

app/src/main/java/emu/skyline/preference/ProfilePicturePreference.kt
```diff
@@ -0,0 +1,83 @@
...
+import androidx.preference.R
+import emu.skyline.SkylineApplication
+import emu.skyline.getPublicFilesDir
+import java.io.File
+import java.io.FileOutputStream
+import java.io.InputStream
+
+class ProfilePicturePreference @JvmOverloads constructor(context : Context, attrs : AttributeSet? = null, defStyleAttr : Int = R.attr.preferenceStyle) : Preference(context, attrs, defStyleAttr) {
+    private val pickMedia = (context as ComponentActivity).registerForActivityResult(ActivityResultContracts.PickVisualMedia()) { uri ->
+        val profilePictureDir = SkylineApplication.instance.getPublicFilesDir().canonicalPath + "/switch/nand/system/save/8000000000000010/su/avators"
```
lynxnb, line 30: Avoid using application context here, do `context.getPublicFilesDir()`.

app/src/main/java/emu/skyline/preference/ProfilePicturePreference.kt
```diff
@@ -0,0 +1,83 @@
...
+        val profilePictureName = "profile_picture.jpeg"
+        try {
+            if (uri != null) { // The user selected a picture
+                PreferenceManager.getDefaultSharedPreferences(context).edit().putString(key, "$profilePictureDir/$profilePictureName").apply()
+                File(profilePictureDir).mkdirs()
+                context.applicationContext.contentResolver.let { contentResolver : ContentResolver ->
```
lynxnb, line 36: Same here, pretty sure you don't need applicationContext, so just do `context.contentResolver`

app/src/main/java/emu/skyline/preference/ProfilePicturePreference.kt
```diff
@@ -0,0 +1,83 @@
...
+                    val readUriPermission : Int = Intent.FLAG_GRANT_READ_URI_PERMISSION
+                    contentResolver.takePersistableUriPermission(uri, readUriPermission)
+                    contentResolver.openInputStream(uri)?.use { inputStream : InputStream ->
+                        var bitmap = BitmapFactory.decodeStream(inputStream)
+                        // Compress the picture
+                        bitmap = Bitmap.createScaledBitmap(bitmap, 256, 256, false)
+                        StoreBitmap(bitmap, "$profilePictureDir/$profilePictureName")
+                    }
```
lynxnb, line 44: Ideally this should be run on a coroutine to avoid jamming the UI thread, since you're compressing a JPEG.

app/src/main/java/emu/skyline/preference/ProfilePicturePreference.kt
```diff
@@ -0,0 +1,83 @@
...
+                }
+            } else { // No picture was selected, clear the profile picture if one was already set
+                if (File("$profilePictureDir/$profilePictureName").exists()) {
+                    File("$profilePictureDir/$profilePictureName").delete()
+                }
+                PreferenceManager.getDefaultSharedPreferences(context).edit().putString(key, "No picture selected").apply()
```
lynxnb, line 50: Avoid using `PreferenceManager` since it's deprecated as of API level 29 (minimum we support is 29). Instead use the `sharedPreferences` member of the `Preference` class.

app/src/main/java/emu/skyline/utils/PreferenceSettings.kt
```diff
@@ -25,6 +25,7 @@ class PreferenceSettings @Inject constructor(@ApplicationContext private val con
     // System
     var isDocked by sharedPreferences(context, true)
     var usernameValue by sharedPreferences(context, context.getString(R.string.username_default))
+    var profilePictureValue by sharedPreferences(context, "")
```
lynxnb, line 28: Name this preference `profilePicturePath` instead of value. Reflect this change on the C++ side as well.

app/src/main/java/emu/skyline/preference/ProfilePicturePreference.kt
lynxnb: Also on the same line, you shouldn't save the string `No picture selected` to shared preferences, but an empty string. See comment below for the `init` block.

app/src/main/java/emu/skyline/preference/ProfilePicturePreference.kt
```diff
@@ -0,0 +1,83 @@
...
+            notifyChanged()
+        } catch (e : Exception) {
+            e.printStackTrace()
+        }
+    }
+
+    init {
+        summaryProvider = SummaryProvider<ProfilePicturePreference> { preference ->
+            Uri.decode(preference.getPersistedString("No picture selected"))
+        }
```
lynxnb, line 61: You should retrieve the value of the preference in this block and check against an empty string. In that case show the `No picture selected` string, otherwise show the actual preference value.

See [GpuDriverPreference.kt](https://github.com/skyline-emu/skyline/blob/ftx1/app/src/main/java/emu/skyline/preference/GpuDriverPreference.kt#L34) for reference.

lynxnb (changes requested):

lynxnb:
Since this PR is already merged, these can be done in a follow-up PR by anyone interested, or I can do them later on.

PR #2081, opened 2022-12-18
Fix small typo

TheASVigilante:
Fixes a small typo with comically large consequences. Fixes textures randomly turning black in Pokemon Let's go games and most likely a lot of other games as well.

TLDR: It was copying from the destination to the source before.


bylaws:
thanks for catching this :)

PR #2109, opened 2022-12-30
Display a preview of the current profile picture

AbandonedCart:
Realized I had no idea how the picture looked after selecting it, so I used the existing preference icon as a preview. 

Also adds a recycle of the bitmap after storing it to free up the resources, since bitmaps tend to be expensive left around.

### DIFF
--- a/app/src/main/java/emu/skyline/preference/ProfilePicturePreference.kt
+++ b/app/src/main/java/emu/skyline/preference/ProfilePicturePreference.kt
@@ -10,6 +10,7 @@ import android.content.Context
 import android.content.Intent
 import android.graphics.Bitmap
 import android.graphics.BitmapFactory
+import android.graphics.drawable.BitmapDrawable
 import android.net.Uri
 import android.util.AttributeSet
 import androidx.activity.ComponentActivity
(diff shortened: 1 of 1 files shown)


AbandonedCart:
Tossed in a couple other fixes, which are explained in the commit message that slightly offset the added cost of generating a preview image.

app/src/main/java/emu/skyline/preference/ProfilePicturePreference.kt
```diff
@@ -59,14 +61,26 @@ class ProfilePicturePreference @JvmOverloads constructor(context : Context, attr
...
+        updatePreview()
     }
 
     override fun onClick() = pickMedia.launch(PickVisualMediaRequest(ActivityResultContracts.PickVisualMedia.ImageOnly))
 
+    private fun updatePreview() {
+        var drawable: BitmapDrawable? = null
+        if (File(profilePicture).exists()) {
+            val preview = BitmapFactory.decodeFile(profilePicture)
+            if (null != preview) {
```
lynxnb, line 73: I find the choice of ordering here quite funny, but we generally would do it like `preview != null`.

lynxnb: Yes, checking for `null` there is completely fine. I just found it funny that `null` came first in the statement.

lynxnb (approved):
LGTM, thanks!

app/src/main/java/emu/skyline/preference/ProfilePicturePreference.kt
AbandonedCart: It's called a yoda condition. A lot of developers use it to avoid accidental assignment. I adapted it back when it was a popular way to solve misleading exceptions in a String.equals() comparison. It never gained enough traction to become the norm for Android, though.

AbandonedCart:
Just out of curiosity, the edge tag on this PR doesn't mean that this PR is only being offered to subscribers does it? I would hope that changes submitted by others aren't being used for that, but it's the same name and color of the subscription tag that requires a $5 / month minimum.

dima-xd:
> Just out of curiosity, the edge tag on this PR doesn't mean that this PR is only being offered to subscribers does it? I would hope that changes submitted by others aren't being used for that, but it's the same name and color of the subscription tag that requires a $5 / month minimum.

The Edge tag means that this PR is currently merged in edge build. It will be merged after 2 weeks of creation date

dima-xd:
> @dima-xd I'm not sure how I feel about outside contributions being used like that. It feels a bit like you are being charged for something I wasn't paid to do, so as a contributor and user, we both get screwed.

Edge is just mainline with all merged PRs. So there is no too much difference. All PRs will be merged anyway, so I don't think it's a big problem

AbandonedCart:
@dima-xd I think you're missing my point. It's probably something that should be a little more obvious here, not Discord or some other third-party resource. It sounds like you went into it knowing. I didn't.

PixelyIon:
@AbandonedCart Edge is generally our way to test commits prior to making their way onto official CI builds so we effectively have a 2-week testing period prior to merging. I can understand the opposition to this and as a result, we've decided to skip this period for this PR (and any other PRs you may have in the future) with CI builds being done ahead of time after the initial review is complete.

AbandonedCart:
I have no problem with it, had I known. There is still work involved in review, merging, writing changelogs, etc and beta subscriptions are nothing new. All I was saying is maybe a little warning would go a long way to prevent anyone from feeling tricked.

AbandonedCart:
"Please note: All submissions are subject to a minimum two week review period, during which time these builds will be available exclusively to edge users to ensure proper testing and quality feedback."

Oh, OK. Cool.

PixelyIon:
That's fair and in that case we'll follow the normal procedure for any future PRs from you. We'll add this to our contributing guidelines so people can know in advance. Regardless, thanks for the contributions.

AbandonedCart:
Sounds good. Thanks for understanding. Just so you know, I'm not trying to give you a hard time. I'd rather say something now than have it bite you guys later.

PR #2119, opened 2023-01-08
Misc gpu fixes

bylaws:
See commit messages

### DIFF
--- a/app/src/main/cpp/skyline/common/circular_queue.h
+++ b/app/src/main/cpp/skyline/common/circular_queue.h
@@ -4,6 +4,7 @@
 #pragma once
 
 #include <common/trace.h>
+#include <common/spin_lock.h>
 #include <common.h>
 
 namespace skyline {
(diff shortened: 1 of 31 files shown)


app/src/main/cpp/skyline/common/circular_queue.h
```diff
@@ -77,46 +78,50 @@ namespace skyline {
...
+                break;
             }
-            produceCondition.notify_one();
+
+        }
+
+        void Append(span<Type> buffer) {
+            for (const auto &item : buffer)
+                Push(item);
         }
```
lynxnb, line 125: Previously `Append` was guaranteed to append elements in `buffer` in a contiguous manner, but that's not the case anymore. I wonder if that's a useful requirement, do we need it at all? E.g. one thread inserts `A,B,C` with `Append` and another one pushes `D` with `Push`, the final order would be `A,B,C,D` before, but now it's not guaranteed.

In any case, a comment stating this explicitly would make sense.

bylaws: That's fine, I don't think that was ever a particular guarantee but ye will mention

PR #2141, opened 2023-01-11
Nvdrv gpu misc

Dearlightgda:
Please do something in the emulator to solve the problems in the non-snapdragon processor, my processor is a kirin710a and I can only open the cuphead to the home screen since when I go to start a game it stays and takes me out of the application
Miren
Screenshot_20230111_131549_skyline emu


PixelyIon:
We are working on better performance for non-Adreno devices. Please don't open PRs to report issues nor should you really report performance as a device issue as we're aware of this already.

PR #2152, opened 2023-01-15
Making skyline compatible with 32bit android phones

gdandsm64fan123:
Hey there! I tried to install Skyline on a 32bit phone but i didnt install, can you make a branch called 32bit? if so, thx!

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -25,7 +25,7 @@ git clone https://github.com/skyline-emu/skyline.git --recursive
 ```
 
 Open Android Studio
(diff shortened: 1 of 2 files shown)


JERD001:
First of all don't open a pull request, that's why there's a issues section. To answer your question, no, Skyline will never support 32bit phones nor Android 32bit. Android phones capable of only 32bit too weak for the task, supposing you have a newer phone that is capable of 64bit but your OS is 32bit you have a low end phone that still isn't good enough, to make it worse best case scenario you have 4gb of RAM and that for some games is just not enough.

PR #2178, opened 2023-02-04
quick and dirty implementation for GetFreeSpaceSize

ErwinSpitaler:
solves the problem that you can't save in some games.
I am still trying to figure out how to implement it correctly

### DIFF
--- a/app/src/main/cpp/skyline/services/fssrv/IFileSystem.cpp
+++ b/app/src/main/cpp/skyline/services/fssrv/IFileSystem.cpp
@@ -85,4 +85,10 @@ namespace skyline::service::fssrv {
     Result IFileSystem::Commit(type::KSession &session, ipc::IpcRequest &request, ipc::IpcResponse &response) {
         return {};
     }
+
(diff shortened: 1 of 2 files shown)


AbandonedCart:
On the Android side, [`https://developer.android.com/reference/android/os/StatFs.html#getAvailableBlocks%28%29](https://developer.android.com/reference/android/os/StatFs.html) has methods to determine the actual storage available. Once you have that, you would be working with real values.

AbandonedCart:
The NDK includes [statvfs](https://chromium.googlesource.com/android_tools/+/20ee6d20/ndk/platforms/android-21/arch-arm/usr/include/sys/statvfs.h), so maybe that SO code should be something like -

```
#include <sys/statvfs.h>

typedef unsigned long long msize;
msize getAvailableDiskSpace() {
    struct statvfs data;
    if (fstatvfs(NULL, &data) < 0)
        return 0;
    return (msize)data.f_bsize * data.f_bavail;
}
```

bylaws:
I've added this to edge for testing, your impl is fine really as VFS will be rewritten anyway.

PR #2180, opened 2023-02-04
Stub indirect layer VI service functions

PixelyIon:
Indirect layers are used by the game to render layer on its own, the game allocates a buffer with the size from `GetIndirectLayerImageRequiredMemoryInfo` and uses `GetIndirectLayerImageMap` to draw the applet contents into the buffer. 

As we don't LLE applet implementations nor do our HLE implementations draw equivalent applets, we cannot submit this to the guest. As a result, these functions are stubbed with the framebuffer being cleared to red. 

Stubbing these functions allows titles such as Dark Souls to proceed further.


bylaws:
markus pls fix

chostdz:
Yeah mark do it :)

PR #2194, opened 2023-02-13
`onBackPressed` -> `onBackPressedDispatcher`

AbandonedCart:
Not a strictly necessary change at the the moment, but Android is removing support for the `onBackPressed` callback to implement perhaps one of their sillier design changes where you can preview where you are going back to before inevitably going back there regardless. 

The way it's explained and the way it functions are the difference between seeing an artist rendering of a new console and then seeing the release product. Ironically, this is the change required for API 13+ and then the actual change to make the new transitions work is a separate condition for API 33 + (that is not currently included in this PR). It almost makes this seem pointless, but I guess it keeps it from colliding with the new handling.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -12,6 +12,7 @@
 
     <application
         android:name=".SkylineApplication"
+        android:enableOnBackInvokedCallback="true"
         android:allowBackup="true"
         android:extractNativeLibs="true"
         android:fullBackupContent="@xml/backup_descriptor"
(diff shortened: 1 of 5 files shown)


AbandonedCart:
Looks like mirroring the target with an `onBackInvokedDispatcher` is unnecessary when using androidx.activity 1.6 or newer and this is already using 1.6.1, so... not even worth the condition.

PixelyIon (approved):
Thanks for doing this, it helps us focus on other things. I've had a cursory glance at it and found nothing wrong but @lynxnb should review as well.

AbandonedCart:
I've been testing it for a couple weeks. I didn't submit it any sooner because it seemed to conflict with exiting the emulation process for a while, but that was because I followed the documentation. `onCreate` only works if you never intend to put the activity in the background because it removes the callback.

lynxnb (approved):
LGMT. I was going to do this at some point, thanks for doing it!

PR #2199, opened 2023-02-14
Add picture in picture for emulation activity

AbandonedCart:
When you launch a game and hit the home or recents button, the emulation activity does picture-in-picture similar to the YouTube app. This one is good for when you have to send a text or something else you need to launch from the home screen and want to make sure the game isn't killed off in the background. The onscreen controls are currently hidden in PiP mode and restored when returning.

It can be completely disabled by either going into PiP and pressing the gear, which opens the PiP settings or (on Samsung and potentially on most Android devices) by opening the Settings app and going to Apps -> Overflow menu (3 dots, top right) -> Special access -> Picture-in-picture.

Not sure if this is something desirable, but I had fun with it.

<details>
  <summary>Screenshot</summary>
  
signal-2023-02-14-112817_002
</details>

### DIFF
--- a/app/src/main/java/emu/skyline/EmulationActivity.kt
+++ b/app/src/main/java/emu/skyline/EmulationActivity.kt
@@ -90,9 +90,10 @@ class EmulationActivity : AppCompatActivity(), SurfaceHolder.Callback, View.OnTo
      */
     private var desiredRefreshRate = 60f
 
-    private val muteIntentAction = "$packageName.EMULATOR_MUTE"
     private lateinit var pictureInPictureParamsBuilder : PictureInPictureParams.Builder
-    private lateinit var muteReceiver : BroadcastReceiver
+    private val pauseIntentAction = "$packageName.EMULATOR_PAUSE"
+    private val muteIntentAction = "$packageName.EMULATOR_MUTE"
+    private lateinit var pictureInPictureReceiver : BroadcastReceiver
 
     @Inject
     lateinit var appSettings : AppSettings
@@ -103,6 +104,8 @@ class EmulationActivity : AppCompatActivity(), SurfaceHolder.Callback, View.OnTo
     lateinit var inputManager : InputManager
 
     lateinit var inputHandler : InputHandler
+
+    private var gameSurface : Surface? = null
 
     /**
      * This is the entry point into the emulation code for libskyline
@@ -273,12 +276,22 @@ class EmulationActivity : AppCompatActivity(), SurfaceHolder.Callback, View.OnTo
 
         pictureInPictureParamsBuilder = PictureInPictureParams.Builder()
 
+        val pictureInPictureActions : MutableList<RemoteAction> = mutableListOf()
+        val pendingFlags = PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE
+
+        val pauseIcon = Icon.createWithResource(this, R.drawable.ic_pause)
+        val pausePendingIntent = PendingIntent.getBroadcast(this, R.drawable.ic_pause, Intent(pauseIntentAction), pendingFlags)
+        val pauseRemoteAction = RemoteAction(pauseIcon, getString(R.string.pause), getString(R.string.pause_emulator), pausePendingIntent)
+        pictureInPictureActions.add(pauseRemoteAction)
+
         if (!emulationSettings.isAudioOutputDisabled) {
             val muteIcon = Icon.createWithResource(this, R.drawable.ic_volume_mute)
-            val mutePendingIntent = PendingIntent.getBroadcast(this, R.drawable.ic_volume_mute, Intent(muteIntentAction), PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE)
+            val mutePendingIntent = PendingIntent.getBroadcast(this, R.drawable.ic_volume_mute, Intent(muteIntentAction), pendingFlags)
             val muteRemoteAction = RemoteAction(muteIcon, getString(R.string.mute), getString(R.string.disable_audio_output), mutePendingIntent)
-            pictureInPictureParamsBuilder.setActions(mutableListOf(muteRemoteAction))
+            pictureInPictureActions.add(muteRemoteAction)
         }
+
+        pictureInPictureParamsBuilder.setActions(pictureInPictureActions)
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S)
             pictureInPictureParamsBuilder.setAutoEnterEnabled(true)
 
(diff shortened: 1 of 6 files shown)


AbandonedCart:
Left some of the Android Studio cleanup in there, but I can strip that out if needed.

bylaws:
This is neat, I like it

lynxnb:
Ok, this is really cool and I've never even thought of adding this to Skyline. Very appreciated!

lynxnb (approved):

app/src/main/java/emu/skyline/EmulationActivity.kt
```diff
@@ -314,6 +338,12 @@ class EmulationActivity : AppCompatActivity(), SurfaceHolder.Callback, View.OnTo
         }
     }
 
+    override fun onUserLeaveHint() {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.S) {
+            enterPictureInPictureMode(PictureInPictureParams.Builder().build())
+        }
```
PixelyIon, line 344:
```suggestion
        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.S)
            enterPictureInPictureMode(PictureInPictureParams.Builder().build())
```

AbandonedCart:
Auto does all the work after Android 12, but I might need the aspect ratio and / or SourceRectHint for Android 11. It definitely needs testing on Android 11.

app/src/main/java/emu/skyline/EmulationActivity.kt
```diff
@@ -225,6 +228,10 @@ class EmulationActivity : AppCompatActivity(), SurfaceHolder.Callback, View.OnTo
             }
         }
 
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+            setPictureInPictureParams(PictureInPictureParams.Builder().setAutoEnterEnabled(true).build())
+        }
```
PixelyIon, line 233:
```suggestion
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S)
            setPictureInPictureParams(PictureInPictureParams.Builder().setAutoEnterEnabled(true).build())
```

PixelyIon (changes requested):
Just two minor style-related nits. Great job though, I had no idea that Android made it that easy to allow PiP. I wonder if this will prevent killing Skyline in lieu of a foreground notification.

app/src/main/java/emu/skyline/EmulationActivity.kt
AbandonedCart: > I wonder if this will prevent killing Skyline in lieu of a foreground notification.

To some extent. As long as the window stays on the screen, it's as if the app is running. If someone kills the PiP, then it's as if they exited the app. Not unlike a notification, but it won't survive sleep (as far as I know).

I should also see if there is a way to add a mute button or at least make one available in the window. I can think of a few cases that may be needed (ie a phone call).

lynxnb:
> I should also see if there is a way to add a mute button or at least make one available in the window. I can think of a few cases that may be needed (ie a phone call).

That sounds good, it probably involves telling it to the c++ side, which you can do by modifying the `nativeSettings` object and then calling `updateNative` on it. (After PR #2037 is merged)

We have a setting to disable audio, that should work.

app/src/main/java/emu/skyline/EmulationActivity.kt
PixelyIon: It might also be possible to add a pause button where you just kill the native surface from Java, that should keep Skyline resident in memory but let it remain pausable at will. While this is probably not the best way to pause things and we'll likely replace it in the future, it does work for everything so far afaik. You can do this by making `Surface?` a member of `EmulationActivity` where it can be assigned to in `surfaceCreated` and null'd in `surfaceDestroyed`, to pause you can simply call `setSurface(null)` and to resume you can call `setSurface(surface)`.

AbandonedCart:
The PiP actions deal in PendingIntent, so it's a regular circus trying to get from point A to point B technically only 2 lines apart in the code.

AbandonedCart:
Sheesh. Android Studio was really struggling with android.view and androidx.core.view

AbandonedCart:
Mute mutes the emulator, but also restores it to the full screen. That's not going to work.

AbandonedCart:
Going through the onNewIntent route leads to restoring the full screen view, so now I'm taking a run at the BroadcastReceiver version of it. If that doesn't work out, then it will be the ugly show and hide a button route.

AbandonedCart:
Mute works, but only appears when the user has not already disabled the audio in settings. It is a window button, so it will look similar to the media controls for the YouTube PiP.

PixelyIon:
Any luck with https://github.com/skyline-emu/skyline/pull/2199#discussion_r1106109737? It's fine if it's not possible or something, just wanted to ensure you didn't miss it.

AbandonedCart:
> Any luck with [#2199 (comment)](https://github.com/skyline-emu/skyline/pull/2199#discussion_r1106109737)? It's fine if it's not possible or something, just wanted to ensure you didn't miss it.

I did miss it. I have to run out quick, but that shouldn't be a problem. I can use the route already available for the mute to add a pause button.

umar248:
If we could get a save state functionality that would be awesome

AbandonedCart:
> If we could get a save state functionality that would be awesome

That's not going to be part of this. Sorry for your loss.

AbandonedCart:
@PixelyIon I think I added it, but I am still building to test that it actually works. Regardless, the framework for adding actions is filled in, so there is some potential there (ie. Once saves do exist, being able to hit save from the window because you realized you need to quit).

AbandonedCart:
Fair enough. Fortunately, the Switch is modern enough that most games have an auto-save feature in the game and the console save is supported. It's not like the old days of Resident Evil where you have to find ribbon, track down a typewriter, and die 3 hours into it and have to start over.

AbandonedCart:
Seems that pause works, except when there is background music (ie. title screens). It will continue to play the loop, so it looks like pause will need to must the audio to be safe.

AbandonedCart:
Alright. Search is failing me. Anyone know offhand where the click action for `on_screen_controller_toggle` is assigned?

AbandonedCart:
FYI - I added the three of you (skyline) to the private repository I made for the sake of transparency. Since I am building stuff that is otherwise restricted to the edge builds, I figured it should be moved off of my public fork.

bylaws:
needs rebaase

bylaws:
```
     Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String android.content.Context.getPackageName()' on a null object reference
        at android.content.ContextWrapper.getPackageName(ContextWrapper.java:183)
        at emu.skyline.EmulationActivity.<init>(EmulationActivity.kt:87)
        at java.lang.Class.newInstance(Native Method)
        at android.app.AppComponentFactory.instantiateActivity(AppComponentFactory.java:95)
        at androidx.core.app.CoreComponentFactory.instantiateActivity(CoreComponentFactory.java:45)
        at **android.app.Instrumentation.newActivity(Instrumentation.java:1329)
        ```
this doesn't work for me

File: app/src/main/java/emu/skyline/EmulationActivity.kt [hunk 3]
```diff
@@ -84,8 +84,8 @@ class EmulationActivity : AppCompatActivity(), SurfaceHolder.Callback, View.OnTo
     private var isEmulatorPaused = false
 
     private lateinit var pictureInPictureParamsBuilder : PictureInPictureParams.Builder
-    private val pauseIntentAction = "$packageName.EMULATOR_PAUSE"
-    private val muteIntentAction = "$packageName.EMULATOR_MUTE"
+    private val pauseIntentAction = "emu.skyline.EMULATOR_PAUSE"
```
bylaws, line 87: i don't think this will work for the dev* variants, maybe you could just latteinit this?

AbandonedCart: A string constant? It's like a request code. Using the package name is just common practice.

AbandonedCart: It is a little misleading, so I updated the names to better match the system ones.

lynxnb:
@AbandonedCart Per-game settings was just merged, would you mind rebasing this on top of `ftx1`? I can do that for you if you'd like, or if you want to do yourself ask away if you encounter any issues.

AbandonedCart:
The last commit was waiting on that change, anyway. It was only a slight dumpster fire because I forgot to plan for each commit having to merge individually.

lynxnb:
Thanks for fixing annoying warnings on `getSerializable` with 2a4c08db91d7d5c0dc7198ac8cd9835062f28dd2, I would have had to fix them myself later otherwise.

app/src/main/java/emu/skyline/EmulationActivity.kt
```diff
@@ -342,6 +393,79 @@ class EmulationActivity : AppCompatActivity(), SurfaceHolder.Callback, View.OnTo
...
+            }.also {
+                registerReceiver(pictureInPictureReceiver, it)
+            }
+
+            binding.onScreenControllerView.isGone = true
+            binding.onScreenControllerToggle.isGone = true
+            binding.onScreenPauseToggle.isGone = true
+        } else {
+            try {
+                if (this::pictureInPictureReceiver.isInitialized)
```
lynxnb, line 448: IMO having `pictureInPictureReceiver` a lateinit var on which you check `isInitialized` is a bad use of lateinit. What I'd suggest doing is declare the receiver like the following, right above the `onPictureInPictureModeChanged` method, outside of it:
```kt
private val pictureInPictureReceiver = object : BroadcastReceiver() {
    override fun onReceive(context : Context, intent : Intent) {
        if (intent.action == intentActionPause)
            pauseEmulator()
        else if (intent.action == intentActionMute)
            changeAudioStatus(false)
    }
}
```

lynxnb (changes requested):

app/src/main/java/emu/skyline/EmulationActivity.kt
AbandonedCart: lateinit in general is a bad use of lateinit. The only purpose it really has is to not deal with ? and !!

lynxnb (approved):
LGTM, good job!

PR #2214, opened 2023-02-23
Implement per-game settings support

lynxnb:
Introducing support for per-games settings. Allows games to have their custom settings applied during emulation. Custom settings are saved to separate `SharedPreferences` files, based on the title ID if present, or a unique enough key otherwise.

The whole settings code has undergone significant rewrite to improve modularity. A `settings` package has been created to accomodate all settings code.
This PR also reorder settings into meaningful categories, a long-standing issue with our settings UI.

For various reasons, the following settings will still use the global values (technical debt is a bitch):
- Log level (will be fixed in #304)
- Profile picture (will be fixed later on, no ETA)
Changing these settings in per-game settings screens will result in values being updated for all games.


AbandonedCart:
Looks good. You may want some sort of note in the main settings that games can be configured individually. I knew to check a game menu, but it may be easy to overlook if you don't know it's there.

lynxnb:
We might include that in a future setup-guide at first boot, not something I'd like take up space in the settings screen.

AbandonedCart:
Having an issue titled "Game-specific settings" that says they are included also works to prevent (some) posting an issue asking if they exist when they're missed.

PR #2222, opened 2023-02-28
Resolve the awkwardness of photo preview

AbandonedCart:
This solves the entire image path being listed, which is excess information because the app's public files path is standardized across all Android apps. While the app needs that information to load the image, it makes for a really long summary.


lynxnb:
Move username and pfp at the bottom of the system category, below system region, instead of at the top.

AbandonedCart:
What's the logic behind that?

AbandonedCart:
Try that out. It moves it below system, but without forcing the picture below the name.

lynxnb:
> What's the logic behind that?
We'd rather have more useful stuff at the top such as docked mode, not username and pfp.

Also I'd argue that a category just for username and pfp is unnecessary, I'd just put at the bottom of the system category.

AbandonedCart:
I can't really get on board with reordering it to the bottom, so I dropped moving it at all for now.

lynxnb (approved):

PR #2230, opened 2023-03-02
Update Kotlin to 1.8.10 + dependencies

nickbeth:
The usual maintenance.


lynxnb:
Kotlin compiler breaks on this one, needs checking.

AbandonedCart:
Might be the version of Android Studio, not the configuration, too.

PR #2232, opened 2023-03-03
Fix depth-stencil formats

bylaws:
These were the opposite to what they should have been.


SlashBlk (commented):
Fixed blurry in Pokemon Briliant Diamond but also removed all supposed depth of view effect.

PR #2236, opened 2023-03-05
Implement usage tracker and basic indirect draw HLE

bylaws:
- Introduces a lot of new gpu debugging infra
- Doubles performance in nier automata thanks to indirect HLE
- Avoids many redundant waits that happened with DMI before due to not being able to tell if resources are dirty
- Fixes DMI on 8 gen 2


AbandonedCart:
I'm not sure how useful this info will be on its own, so let me know what you need and I will get it. When applying this PR, cutscenes have a red hue to them. The best example would be Sonic Frontiers because it's visible right in the opening.

PR #2246, opened 2023-03-09
Migrate to Material Design 3

lynxnb:
The app has been migrated to Material Design 3. Additionally, Material You support has been added.


AbandonedCart:
This solves the profile icon looking out of place without moving it, too. I've never been a fan of the preference icon being a thing, but it's also not really my place to strip it all out over one preview. Good choice.

bylaws:
pls rebase

PR #2247, opened 2023-03-09
Support  deprecation of `getSerializableExtra`

AbandonedCart:
Split this out from #2199 because of the merge conflict it would have with #2246 since it has nothing to do with picture in picture and is easier to fix alone.


lynxnb (approved):
Thanks for splitting this into its own PR!

Issue #2218, opened 2023-02-26
Needs UI Friendly way to Remove / Delete NSP / XCI Games/Software/DLC/Updates from Skyline Game Library Menu 


RFL0WGitPR:
Skyline Emulator could use a UI Friendly way to manage storage data space on installed Games on Skyline Games Library Menu

The software and icon on the HOME Menu will be deleted to manage storage space.

This feature should remove all Game Data /Software Apps/ Screenshots, or videos for related title when user selects / press and holds a Game title on App.

User Presses and Holds Selected Game [contextual Pop Up menu appears] with management options such as Delete

Feature should work with both Touch / Pro Controller / 3rd party Bluetooth controller

AbandonedCart:
Is there some screenshot / video feature I missed? I know the Switch has a share button, but I didn't think the emulator had one specific to the emulator. If it doesn't, it means the screenshots and videos you made are being done through Android and that's not stuff that skyline, as an individual app, has the sort of access to determine belongs to a specific game.

PR #2252, opened 2023-03-12
Add a delete button to the game dialog

AbandonedCart:
Bear with me because there is some stuff that would normally be unnecessary, but needs to be included because of timing.

This adds the ability to delete a file through the game dialog. The problem, as you may already know, is that write permission is not being requested to display the game list. That is why this has to check for write permission to be present and then call a new instance of selecting your root folder to add it. By doing it this way and not simply trying to shoehorn write permission on top of the existing request, it prevents ever having to request it again for subsequent files. As for new requests (ie. new installs), the permission will be requested as part of the initial setup. I'm sure there are 100 prettier ways to go about this, but most would involve having requested the write permission up front, despite not needing it yet.

Oh, and the delete button is intentionally the negative button to prevent accidental presses. Anyone rushing through dialogs should end up cancelling, not deleting files.

### DIFF
--- a/app/src/main/java/emu/skyline/AppDialog.kt
+++ b/app/src/main/java/emu/skyline/AppDialog.kt
@@ -105,6 +105,11 @@ class AppDialog : BottomSheetDialogFragment() {
             shortcutManager.requestPinShortcut(info.build(), null)
         }
 
+        binding.gameDelete.setOnClickListener {
+            dialog?.onBackPressed()
+            (activity as MainActivity).requestAppItemDelete(item)
+        }
+
         binding.gameTitleId.setOnLongClickListener {
             val clipboard = requireActivity().getSystemService(android.content.Context.CLIPBOARD_SERVICE) as android.content.ClipboardManager
             clipboard.setPrimaryClip(android.content.ClipData.newPlainText("Title ID", item.titleId))
--- a/app/src/main/java/emu/skyline/AppDialog.kt
+++ b/app/src/main/java/emu/skyline/AppDialog.kt
@@ -107,7 +107,7 @@ class AppDialog : BottomSheetDialogFragment() {
 
         binding.gameDelete.setOnClickListener {
             dialog?.onBackPressed()
-            (activity as MainActivity).requestAppItemDelete(item)
+            (activity as MainActivity).requestDeleteAppItem(item)
         }
 
         binding.gameTitleId.setOnLongClickListener {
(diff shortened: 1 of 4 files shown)


Ragul32:
Add screenshot with comments it will be useful

AbandonedCart:
> Add screenshot with comments it will be useful

Of the delete button or the file being there and then not?

Ragul32:
> > Add screenshot with comments it will be useful
> 
> Of the delete button or the file being there and then not?

Like showing delete button with dialog box that's all

AbandonedCart:
> Like showing delete button with dialog box that's all

They aren't displayed at the same time.

<details>
<summary>Delete button</summary>

delete
</details>

<details>
<summary>Delete dialog</summary>

dialog
</details>

As you can see, they are pretty straightforward.

lynxnb:
I can't see any valid use-case for this honestly. Deleting a game would be something we can easily delegate to file manager, which doesn't need UI elements on our end. Did you have any specific need that brought you to implement this?

AbandonedCart:
> Did you have any specific need that brought you to implement this?

https://github.com/skyline-emu/skyline/issues/2218

As it stands, you open the emulator, play a game, decide you don't want it, quit the emulator, open a file manager, find the file, and delete the file.

With the basic functionality there, it can also be wired to the home screen icon so when you decide to remove a game, you actually have the option to delete the entire game and not delete the icon, open a file manager, find the file, and delete the file.

Seems like delegating it to a file manager would be a way to avoid the need to write the code for it, but I did that.

AbandonedCart:
The home icon will be a circus, by the way. It would give this more context, but it's one of those things where only system apps have really used it and the documentation is more or less "make an xml and click your heels" but then nothing happens. I have surgery in the morning tomorrow, so hopefully I can put something together after that.

lynxnb:
As I understand #2218 was about having an icon to delete saves/pipeline cache/whatnot in the future, not about deleting the game ROM. There's no reason to expose an option to delete game ROMs in the emulator, for this reason I'm going to close this.

AbandonedCart:
It does make removing old games a lot easier, but it's also perfectly acceptable to _prefer_ not having the option.

As for 2218, it read like it was about deleting **everything**. The game, any shortcuts to it, the cache, saves, and anything else even remotely related. Some of it is flat out impossible because you can't delete, only disable, shortcuts created through the pin method and media captured outside the app only, at best, is named for the app in which it was taken.

Attaching the cache to removing a game from the library is probably the only realistic option left, if it isn't removed already. I don't know about you, but if deleting a game to free up space removed my save, I'd be pissed. Settings can already be cleared, but they aren't significant enough to justify removing them and would likely just need to be set the same way if the game ever returned.

PR #2253, opened 2023-03-12
Add automatic refreshing games list

### LINKED ISSUE #2015: Add automatic refreshing games list
It will be great if the app update games list that are showing like when i add new game to the folder it will be amazing if it show up automatically without scrolling down .

KikiManjaro:
Closes #2015 

Add file watcher (custom one cause `FileObservers` does not work with `DocumentFile`) to game directory and update game ui when needed

### DIFF
--- a/app/src/main/java/emu/skyline/MainActivity.kt
+++ b/app/src/main/java/emu/skyline/MainActivity.kt
@@ -41,6 +41,7 @@ import emu.skyline.provider.DocumentsProvider
 import emu.skyline.settings.AppSettings
 import emu.skyline.settings.EmulationSettings
 import emu.skyline.settings.SettingsActivity
+import emu.skyline.utils.DocumentFileWatcher
 import emu.skyline.utils.GpuDriverHelper
 import emu.skyline.utils.WindowInsetsHelper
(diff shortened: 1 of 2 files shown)


bylaws:
I don't think this is the best way to go about this, ideally the game list should auto update between launches. Reloading at runtime isn't particularly useful

AbandonedCart:
It actually doesn't refresh on launch, whether or not that was intended to happen. I think the code, if it follows a file observer, tracks changes made during runtime. It would only be something like adding a file in the background.

AbandonedCart:
... the second I saw that thread.sleep command I was waiting for that one.

Ever consider a database, similar to how MediaStore handles images? You could basically cache everything about the game once and not keep reading it, caching it somewhere, etc file uri changes. You'd probably remove a lot of existing overhead instead of adding any at all.

KikiManjaro:
Nice i see, i'll try to make something like this :) I may have gone a bit far with this one

lynxnb:
Since we won't be accepting the feature implemented this way, I'm closing this PR to avoid confusion with people thinking this feature is coming. Feel free to either reopen this or open a new PR in the future.

PR #2254, opened 2023-03-12
Add search to settings

PabloG02:
Both categories and preferences are searchable, and queries can be concatenated with `,`.

### DIFF
--- a/app/src/main/java/emu/skyline/settings/SettingsActivity.kt
+++ b/app/src/main/java/emu/skyline/settings/SettingsActivity.kt
@@ -24,6 +24,7 @@ import androidx.preference.PreferenceFragmentCompat
 import androidx.preference.forEach
 import com.google.android.material.appbar.AppBarLayout
 import com.google.android.material.internal.ToolbarUtils
+import emu.skyline.BuildConfig
 import emu.skyline.R
 import emu.skyline.data.AppItemTag
 import emu.skyline.databinding.SettingsActivityBinding
@@ -37,7 +38,11 @@ private const val PREFERENCE_DIALOG_FRAGMENT_TAG = "androidx.preference.Preferen
 
 class SettingsActivity : AppCompatActivity(), PreferenceFragmentCompat.OnPreferenceDisplayDialogCallback {
     val binding by lazy { SettingsActivityBinding.inflate(layoutInflater) }
-    val creditsCategories = arrayOf("category_credits", "category_licenses")
+    val hiddenCategoriesFromSearch = if (BuildConfig.BUILD_TYPE == "release") {
+        arrayOf("category_debug", "category_credits", "category_licenses")
+    } else {
+        arrayOf("category_credits", "category_licenses")
+    }
 
     /**
      * The instance of [PreferenceFragmentCompat] that is shown inside [R.id.settings]
@@ -141,7 +146,7 @@ class SettingsActivity : AppCompatActivity(), PreferenceFragmentCompat.OnPrefere
                 val queries = newText.split(",")
                 if (newText.isNotEmpty()) {
                     preferenceFragment.preferenceScreen.forEach { preferenceCategory ->
-                        if (creditsCategories.contains(preferenceCategory.key)) {
+                        if (hiddenCategoriesFromSearch.contains(preferenceCategory.key)) {
                             preferenceCategory.isVisible = false
                             return@forEach
                         }
--- a/app/src/main/java/emu/skyline/settings/SettingsActivity.kt
+++ b/app/src/main/java/emu/skyline/settings/SettingsActivity.kt
@@ -9,15 +9,20 @@ import android.annotation.SuppressLint
 import android.os.Bundle
 import android.text.TextUtils
 import android.view.KeyEvent
+import android.view.Menu
 import android.view.ViewTreeObserver
 import android.widget.TextView
 import androidx.appcompat.app.AppCompatActivity
+import androidx.appcompat.widget.SearchView
 import androidx.coordinatorlayout.widget.CoordinatorLayout
 import androidx.core.view.WindowCompat
 import androidx.preference.EditTextPreference
 import androidx.preference.ListPreference
 import androidx.preference.Preference
+import androidx.preference.PreferenceCategory
 import androidx.preference.PreferenceFragmentCompat
+import androidx.preference.forEach
+import com.google.android.material.appbar.AppBarLayout
 import com.google.android.material.internal.ToolbarUtils
 import emu.skyline.R
 import emu.skyline.data.AppItemTag
@@ -32,6 +37,7 @@ private const val PREFERENCE_DIALOG_FRAGMENT_TAG = "androidx.preference.Preferen
 
 class SettingsActivity : AppCompatActivity(), PreferenceFragmentCompat.OnPreferenceDisplayDialogCallback {
     val binding by lazy { SettingsActivityBinding.inflate(layoutInflater) }
+    val creditsCategories = arrayOf("category_credits", "category_licenses")
 
     /**
      * The instance of [PreferenceFragmentCompat] that is shown inside [R.id.settings]
@@ -97,6 +103,7 @@ class SettingsActivity : AppCompatActivity(), PreferenceFragmentCompat.OnPrefere
 
         @SuppressLint("RestrictedApi")
         val toolbarTitleTextView = ToolbarUtils.getTitleTextView(binding.titlebar.toolbar)
+
         @SuppressLint("RestrictedApi")
         val toolbarSubtitleTextView = ToolbarUtils.getSubtitleTextView(binding.titlebar.toolbar)
         toolbarTitleTextView?.let { enableMarquee(it) }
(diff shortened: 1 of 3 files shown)


app/src/main/java/emu/skyline/settings/SettingsActivity.kt
```diff
@@ -102,6 +109,69 @@ class SettingsActivity : AppCompatActivity() {
...
+            var noKey = true
+            (preferenceCategory as PreferenceCategory).forEach { preference ->
+                if (preference.key != null) {
+                    mapKeyTitle[preference.title.toString()] = preference
+                    noKey = false
+                }
+            }
+            if (!noKey)
+                mapKeyTitle[preferenceCategory.title.toString()] = preferenceCategory
+        }
```
lynxnb, line 123: This should definitely not be done here every time the menu is inflated, but after the preference screen has been created.

app/src/main/res/menu/menu.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
```
lynxnb, line 1: Give a better name to this file, `settings_menu.xml`.

app/src/main/java/emu/skyline/settings/SettingsActivity.kt
```diff
@@ -23,6 +29,7 @@ import emu.skyline.utils.WindowInsetsHelper
 
 class SettingsActivity : AppCompatActivity() {
     val binding by lazy { SettingsActivityBinding.inflate(layoutInflater) }
+    var mapKeyTitle = mutableMapOf<String, Preference>()
```
lynxnb, line 32: You usually want variables to have meaningful names, in this case `titleToPreferenceMap`. Add a KDoc comment on top to explain why this exists.

app/src/main/java/emu/skyline/settings/SettingsActivity.kt
```diff
@@ -102,6 +109,69 @@ class SettingsActivity : AppCompatActivity() {
...
+                        preferenceCategory.isVisible = !allHiden || filteredMap.containsValue(preferenceCategory)
+                    }
+                } else { // If user input is empty or there's no match, show all preferences (including those without key)
+                    (supportFragmentManager.findFragmentById(R.id.settings) as PreferenceFragmentCompat).preferenceScreen.forEach { preferenceCategory ->
+                        preferenceCategory.isVisible = true
+                        (preferenceCategory as PreferenceCategory).forEach {
+                            it.isVisible = true
+                        }
+                    }
+                }
```
lynxnb, line 168: In the case the input text is empty, all settings should be shown.
In the case the input text doesn't match anything, no preferences should be shown.

app/src/main/java/emu/skyline/settings/SettingsActivity.kt
```diff
@@ -102,6 +109,69 @@ class SettingsActivity : AppCompatActivity() {
...
+
+            override fun onQueryTextChange(newText : String?) : Boolean {
+                val querys = newText!!.split(",").toTypedArray()
+                val filteredMap = mutableMapOf<String, Preference>()
+                for (query in querys) {
+                    filteredMap += mapKeyTitle.filterKeys { it.contains(query as CharSequence, true) }
+                }
+                if (newText.isNotEmpty() && filteredMap.isNotEmpty()) {
+                    (supportFragmentManager.findFragmentById(R.id.settings) as PreferenceFragmentCompat).preferenceScreen.forEach { preferenceCategory ->
+                        var allHiden = true
```
lynxnb, line 151: Name of this should be `areAllPrefsHidden`, with a comment on top saying `// Tracks whether all preferences under this category are hidden`.

app/src/main/java/emu/skyline/settings/SettingsActivity.kt
```diff
@@ -102,6 +109,69 @@ class SettingsActivity : AppCompatActivity() {
...
+            }
+
+            override fun onQueryTextChange(newText : String?) : Boolean {
+                val querys = newText!!.split(",").toTypedArray()
+                val filteredMap = mutableMapOf<String, Preference>()
+                for (query in querys) {
+                    filteredMap += mapKeyTitle.filterKeys { it.contains(query as CharSequence, true) }
+                }
+                if (newText.isNotEmpty() && filteredMap.isNotEmpty()) {
+                    (supportFragmentManager.findFragmentById(R.id.settings) as PreferenceFragmentCompat).preferenceScreen.forEach { preferenceCategory ->
```
lynxnb, line 150: You don't need to `(supportFragmentManager.findFragmentById(R.id.settings) as PreferenceFragmentCompat)`, the current fragment is already `preferenceFragment` member. Fix all instances of this.

app/src/main/java/emu/skyline/settings/SettingsActivity.kt
```diff
@@ -102,6 +109,69 @@ class SettingsActivity : AppCompatActivity() {
...
+                        var allHiden = true
+                        (preferenceCategory as PreferenceCategory).forEach { preference ->
+                            preference.isVisible = filteredMap.containsValue(preference) || filteredMap.containsValue(preferenceCategory)
+                            if (preference.isVisible && allHiden) {
+                                allHiden = false
+                            }
```
lynxnb, line 156: Single line `if`, drop `{}`.

app/src/main/java/emu/skyline/settings/SettingsActivity.kt
```diff
@@ -102,6 +109,69 @@ class SettingsActivity : AppCompatActivity() {
...
+            override fun onQueryTextSubmit(query : String?) : Boolean {
+                return false
+            }
+
+            override fun onQueryTextChange(newText : String?) : Boolean {
+                val querys = newText!!.split(",").toTypedArray()
+                val filteredMap = mutableMapOf<String, Preference>()
+                for (query in querys) {
+                    filteredMap += mapKeyTitle.filterKeys { it.contains(query as CharSequence, true) }
+                }
```
lynxnb, line 148: Single line `for`, drop `{}` and an empty newline below.

app/src/main/java/emu/skyline/settings/SettingsActivity.kt
```diff
@@ -102,6 +109,69 @@ class SettingsActivity : AppCompatActivity() {
...
+                    AppBarLayout.LayoutParams.SCROLL_FLAG_SCROLL
+        }
+
+        searchView.setOnQueryTextListener(object : SearchView.OnQueryTextListener {
+            override fun onQueryTextSubmit(query : String?) : Boolean {
+                return false
+            }
+
+            override fun onQueryTextChange(newText : String?) : Boolean {
+                val querys = newText!!.split(",").toTypedArray()
```
lynxnb, line 144: If `newText` is null you should just return here, and not crash. Look if you can make `newText` non-nullable too.

lynxnb (changes requested):
Some issues here and there I pointed out in the code, but there's a fundamental issue. I don't really understand what's the benefit of the `mapKeyTitle` you introduced. Usually you'd introduce a map to help computation and avoid unnecessary searching, but the way you're using it is not useful at all and only adds to the complexity. You effectively end up doing a linear search over all `Preferences` **and** pay for the map lookup cost.

There's two ways of fixing this:
- Use a linear search algo, get rid of the unnecesary map.
- Actually use a proper data structure to speed up search. In this particular case, a map is not enough as you have the added complexity of having to hide a category if all of its children are hidden. Flattening the natural tree structure of preferences poses no benefit here. I suppose a tree-like structure with custom objects tracking whether all children are hidden could work.

If you don't feel confident enough, I'd suggest you just go with a plain linear search for now. It's fast enough for our current needs anyway.

PabloG02:
In the end I decided to go for the lineal search approach. Hope that it's better now.

app/src/main/java/emu/skyline/settings/SettingsActivity.kt
```diff
@@ -102,6 +107,57 @@ class SettingsActivity : AppCompatActivity() {
...
+        }
+
+        searchView.setOnQueryTextListener(object : SearchView.OnQueryTextListener {
+            override fun onQueryTextSubmit(query : String) : Boolean {
+                return false
+            }
+
+            override fun onQueryTextChange(newText : String) : Boolean {
+                val queries = newText.split(",").toTypedArray()
+                if (newText.isNotEmpty()) {
```
lynxnb, line 131: There's an issue here when the user only inputs `,`, no setting will show up. Ideally when that happens, all settings should show up as the user is searching for "empty strings", which occur in every string. By handling this correctly, you also avoid the need to check if `newText` is empty. Here's how to fix:
```suggestion
                val queries = newText.split(",").toTypedArray()
```

Also you don't need a `TypedArray` here.

app/src/main/java/emu/skyline/settings/SettingsActivity.kt
```diff
@@ -102,6 +107,57 @@ class SettingsActivity : AppCompatActivity() {
...
+                        preferenceCategory.isVisible = hasNonNullPreferenceKey && (!areAllPrefsHidden || categoryVisible)
+                    }
+                } else { // If user input is empty, show all preferences
+                    preferenceFragment.preferenceScreen.forEach { preferenceCategory ->
+                        preferenceCategory.isVisible = true
+                        (preferenceCategory as PreferenceCategory).forEach { preference ->
+                            preference.isVisible = true
+                        }
+                    }
+                }
```
lynxnb, line 154: Drop this `else` block as the case is already handled by the rest of the code if you apply my suggestion above.

lynxnb (changes requested):

app/src/main/java/emu/skyline/settings/SettingsActivity.kt
```diff
@@ -102,6 +107,57 @@ class SettingsActivity : AppCompatActivity() {
...
+
+            override fun onQueryTextChange(newText : String) : Boolean {
+                val queries = newText.split(",").toTypedArray()
+                if (newText.isNotEmpty()) {
+                    preferenceFragment.preferenceScreen.forEach { preferenceCategory ->
+                        val categoryVisible = queries.any { preferenceCategory.title?.contains(it, true) ?: false }
+                        // Tracks whether all preferences under this category are hidden
+                        var areAllPrefsHidden = true
+                        // Tracks whether there is at least one preference under this category that has a key
+                        var hasNonNullPreferenceKey = false
```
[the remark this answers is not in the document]
PabloG02: The way I see it, "preferences" like Edge supporters and licenses should not appear as a search result as it would clutter the results, so the way I have to exclude them is this one, as they are entries that do not  have keys.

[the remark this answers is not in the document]
AbandonedCart: Couldn't you filter them out ahead of time by checking if preferenceCategory has a key of category_credits or category_licenses?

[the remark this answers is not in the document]
PabloG02: I could, but I would prefer not to hardcode stuff if possible. If settings ever change in the future again, that way the search will keep working.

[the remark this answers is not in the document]
AbandonedCart:
```
val documentationKeys = arrayOf("category_credits", "category_licenses")

override fun onCreateOptionsMenu(menu : Menu?) : Boolean {
...
preferenceFragment.preferenceScreen.forEach { preferenceCategory ->
    if (documentationKeys.contains(preferenceCategory.key)) {
        preferenceCategory.isVisible = false
        return@forEach
    }
...
```

It's hard-coded, sure, but also skips a lot of checking that each key exists and it's easy enough to maintain as keys change, get added, get deleted, etc.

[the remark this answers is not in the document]
lynxnb: @AbandonedCart your example looks good, we're going with that

app/src/main/java/emu/skyline/settings/SettingsActivity.kt
```diff
@@ -133,18 +133,18 @@ class SettingsActivity : AppCompatActivity() {
...
-                            // Tracks whether all preferences under this category are hidden
-                            var areAllPrefsHidden = true
-                            (preferenceCategory as PreferenceCategory).forEach { preference ->
-                                preference.isVisible = queryMatchesCategory || queries.any { preference.title?.contains(it, true) ?: false }
-                                if (preference.isVisible && areAllPrefsHidden)
-                                    areAllPrefsHidden = false
-                            }
-                            // Hide PreferenceCategory if none of its preferences match the search and neither the category title
-                            preferenceCategory.isVisible = !areAllPrefsHidden || queryMatchesCategory
+                            return@forEach
```
AbandonedCart, line 136: I only suggested the return because I hate keeping track of brackets and indentation. It's more laziness than anything.

AbandonedCart: Honestly, that makes two of us. I had searched if there was a way to continue a for each loop right before posting it.

lynxnb (approved):
LGTM now, thanks!

PR #2255, opened 2023-03-12
Add xml for the shader compilation screen

PabloG02:
Once implemented, it should look like this:

image

### DIFF
--- a/app/src/main/res/layout/shader_loading.xml
+++ b/app/src/main/res/layout/shader_loading.xml
@@ -65,7 +65,7 @@
                 android:ellipsize="marquee"
                 android:singleLine="true"
                 android:textAlignment="viewStart"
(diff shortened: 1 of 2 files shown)


lynxnb:
There's a simpler way I would implement the layout, and that is the following:
A vertical `LinearLayout` for game name and version, and another one for `Compiling...` text, progress bar and text. The two linear layouts are then positioned vertically either via `ContraintLayout` chains, with a `space-around` policy. A `FlexLayout` would also work. This gets rid of the two nested `ContraintLayout`, which is considered bad practice and performs worse in general.

artslay:
Helpful

lynxnb:
I haven't had time to review this yet, I'll get to it as soon as I can.

nickbeth (approved):

PR #2259, opened 2023-03-14
One missing Bitmap to rule them all

AbandonedCart:
`... and one condition to find them without the need for multiple instances and passing a condition. The data class can simply return the same instance every time it's necessary.`

Jokes aside, there are a minimum of 3 instances of the missingIcon Bitmap per run and it's a static icon. If the explanation isn't convincing enough, compare the load times.


lynxnb (approved):
LGTM, but maybe reword the commit title to something more useful long term. We like funny stuff in our commit descriptions, not titles :)

AbandonedCart:
Boring version, done.

PR #2261, opened 2023-03-15
Don't load Bitmap already set as Drawable

AbandonedCart:
I didn't catch this before, but the ImageView for these two layouts already has the default icon. There is no point checking for null and then replacing the default with the default when the load can be skipped altogether.

### DIFF
--- a/app/src/main/java/emu/skyline/AppDialog.kt
+++ b/app/src/main/java/emu/skyline/AppDialog.kt
@@ -63,7 +63,7 @@ class AppDialog : BottomSheetDialogFragment() {
             }
         }
 
-        binding.gameIcon.setImageBitmap(item.bitmapIcon)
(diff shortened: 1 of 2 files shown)


lynxnb:
AppViewItem layouts have the default bitmap defined as `tool:src` which is only displayed in the layout preview.

https://github.com/skyline-emu/skyline/blob/ftx1/app/src/main/res/layout/app_item_grid.xml#L26

AbandonedCart:
I forgot to turn on split view and assumed I just overlooked it. I imagine more games would actually have icons, which would make fewer instances of applying the default bitmap than not. I don't know of any inherent performance boost for having a default image load in the layout unless it defines the dimensions, but most of those are explicit. I don't think there's an actual improvement to be made here

PR #2271, opened 2023-03-19
Add an option to filter valid files for library

AbandonedCart:
I don't know if this is the case for everyone, but I store DLC and stuff alongside my games in a folder named after the game. This means that when I load the emulator, my list of items is like 30 long, but only 6 of them are games. Filtering the files shouldn't be enforced, but having the option would be nice. For now, it's based on a version being populated, since none of the DLC are actually assigned versions read by the emulator.

### DIFF
--- a/app/src/main/java/emu/skyline/MainActivity.kt
+++ b/app/src/main/java/emu/skyline/MainActivity.kt
@@ -305,6 +305,7 @@ class MainActivity : AppCompatActivity() {
 
     private fun populateAdapter() {
         val items = getDataItems()
+        if (appSettings.filterValid) items.removeIf { item -> item is AppItem && item.version.isNullOrEmpty() }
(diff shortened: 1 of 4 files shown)


AbandonedCart:
Yeah, so changing the name of the preference mid commit wasn't the best idea...

AbandonedCart:
<details>
<summary>Filter Off</summary>

filter_off
</details>

<details>
<summary>Filter On</summary>

filter_on
</details>

bylaws:
idk if i'm a fan of this, it feels like something better solved on a user side, @lynxnb thoughts?

AbandonedCart:
Oh, and to be fair, I'm not fond of it. It feels a bit hacky. I'm not denying that at all.

AbandonedCart:
Closing in favor of a more rational alternative

PR #2260, opened 2023-03-15
Reimplement audren and audout using yuzu's audio_core

bylaws:

PR #2273, opened 2023-03-21
Conditional Rendering and Query support

bylaws:


AbandonedCart:
Looks like this needs a rebase, since the first 6 commits are essentially #2260

PR #2278, opened 2023-03-22
On-Screen Controls Improvements Part 1

lynxnb:
Some long standing issues in on-screen controls have been fixed in this first wave of improvements, with more to come in following PRs.

- Buttons can be aligned to grid
- Button size can be adjusted invididually
- Better default position (you might need to reset buttons to default to see it)
- Separate L3/R3 buttons, disabled by default
- Toned down haptic feedback (vibration) a bit

See individual commits for details.


AbandonedCart:
Do you think it's possible to give controller settings higher priority in the settings list? From a user standpoint, I always feel like I am scrolling past a lot of stuff I never change to get there.

PR #2297, opened 2023-04-07
Import/Export Game Saves

PabloG02:

### DIFF
--- a/app/src/main/java/emu/skyline/AppDialog.kt
+++ b/app/src/main/java/emu/skyline/AppDialog.kt
@@ -28,7 +28,12 @@ import emu.skyline.databinding.AppDialogBinding
 import emu.skyline.loader.LoaderResult
 import emu.skyline.provider.DocumentsProvider
 import emu.skyline.settings.SettingsActivity
+import emu.skyline.utils.ZipUtils
 import emu.skyline.utils.serializable
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import java.io.BufferedOutputStream
 import java.io.File
 import java.io.FileOutputStream
@@ -64,9 +69,13 @@ class AppDialog : BottomSheetDialogFragment() {
                 val saveFolder = File(savesFolderRoot + item.titleId)
                 val inputZip = requireContext().contentResolver.openInputStream(uri)
                 if (inputZip != null) {
-                    emu.skyline.utils.ZipUtils.Companion.unzip(inputZip, saveFolder)
-                    binding.deleteSave.isEnabled = true
-                    binding.exportSave.isEnabled = true
+                    CoroutineScope(Dispatchers.IO).launch {
+                        ZipUtils.unzip(inputZip, saveFolder)
+                        withContext(Dispatchers.Main){
+                            binding.deleteSave.isEnabled = true
+                            binding.exportSave.isEnabled = true
+                        }
+                    }
                 }
             } else {
                 Snackbar.make(binding.root, "Zip file must have as name the TitleID of the game", Snackbar.LENGTH_LONG).show()
@@ -152,30 +161,32 @@ class AppDialog : BottomSheetDialogFragment() {
 
         binding.exportSave.isEnabled = saveExists
         binding.exportSave.setOnClickListener {
-            val saveFolder = File(saveFolderPath)
-            //val outputZipFile = File.createTempFile("out", ".zip")
-            val outputZipFile = File("$saveFolderPath.zip")
-            if (outputZipFile.exists()) outputZipFile.delete()
-            outputZipFile.createNewFile()
-            ZipOutputStream(BufferedOutputStream(FileOutputStream(outputZipFile))).use { zos ->
-                saveFolder.walkTopDown().forEach { file ->
-                    val zipFileName = file.absolutePath.removePrefix(saveFolder.absolutePath).removePrefix("/")
-                    if (zipFileName == "") return@forEach
-                    val entry = ZipEntry("$zipFileName${(if (file.isDirectory) "/" else "")}")
-                    zos.putNextEntry(entry)
-                    if (file.isFile) {
-                        file.inputStream().use { fis -> fis.copyTo(zos) }
+            CoroutineScope(Dispatchers.IO).launch {
+                val saveFolder = File(saveFolderPath)
+                //val outputZipFile = File.createTempFile("out", ".zip")
+                val outputZipFile = File("$saveFolderPath.zip")
+                if (outputZipFile.exists()) outputZipFile.delete()
+                outputZipFile.createNewFile()
+                ZipOutputStream(BufferedOutputStream(FileOutputStream(outputZipFile))).use { zos ->
+                    saveFolder.walkTopDown().forEach { file ->
+                        val zipFileName = file.absolutePath.removePrefix(saveFolder.absolutePath).removePrefix("/")
+                        if (zipFileName == "") return@forEach
+                        val entry = ZipEntry("$zipFileName${(if (file.isDirectory) "/" else "")}")
+                        zos.putNextEntry(entry)
+                        if (file.isFile) {
+                            file.inputStream().use { fis -> fis.copyTo(zos) }
+                        }
                     }
                 }
+                withContext(Dispatchers.Main) {
+                    val file = DocumentFile.fromSingleUri(requireContext(), DocumentsContract.buildDocumentUri(DocumentsProvider.AUTHORITY, "${DocumentsProvider.ROOT_ID}/switch/nand/user/save/0000000000000000/00000000000000000000000000000001/${item.titleId}.zip"))!!
+                    val intent = Intent(Intent.ACTION_SEND)
+                        .setDataAndType(file.uri, "application/zip")
+                        .addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)
+                        .putExtra(Intent.EXTRA_STREAM, file.uri)
+                    startActivity(Intent.createChooser(intent, "Share save file"))
+                }
             }
-
-            val file = DocumentFile.fromSingleUri(requireContext(), DocumentsContract.buildDocumentUri(DocumentsProvider.AUTHORITY, "${DocumentsProvider.ROOT_ID}/switch/nand/user/save/0000000000000000/00000000000000000000000000000001/${item.titleId}.zip"))!!
-            val intent = Intent(Intent.ACTION_SEND)
-                .setDataAndType(file.uri, "application/zip")
-                .addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)
-                .putExtra(Intent.EXTRA_STREAM, file.uri)
-            startActivity(Intent.createChooser(intent, "Share save file"))
-            //outputZipFile.deleteOnExit()
         }
 
         binding.gameTitleId.setOnLongClickListener {
--- a/app/src/main/java/emu/skyline/AppDialog.kt
+++ b/app/src/main/java/emu/skyline/AppDialog.kt
@@ -11,10 +11,13 @@ import android.content.pm.ShortcutInfo
 import android.content.pm.ShortcutManager
 import android.graphics.drawable.Icon
 import android.os.Bundle
+import android.provider.DocumentsContract
 import android.view.KeyEvent
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import androidx.activity.result.contract.ActivityResultContracts
+import androidx.documentfile.provider.DocumentFile
 import com.google.android.material.bottomsheet.BottomSheetBehavior
 import com.google.android.material.bottomsheet.BottomSheetDialogFragment
 import com.google.android.material.snackbar.Snackbar
@@ -22,8 +25,14 @@ import emu.skyline.data.AppItem
 import emu.skyline.data.AppItemTag
 import emu.skyline.databinding.AppDialogBinding
 import emu.skyline.loader.LoaderResult
+import emu.skyline.provider.DocumentsProvider
 import emu.skyline.settings.SettingsActivity
 import emu.skyline.utils.serializable
+import java.io.BufferedOutputStream
+import java.io.File
+import java.io.FileOutputStream
+import java.util.zip.ZipEntry
+import java.util.zip.ZipOutputStream
 
 /**
  * This dialog is used to show extra game metadata and provide extra options such as pinning the game to the home screen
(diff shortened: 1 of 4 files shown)


PabloG02:
If the zip function is OK, I can move it to ZipUtils. Also, not sure if the way I'm using coroutines is the correct one.

app/src/main/java/emu/skyline/AppDialog.kt
```diff
@@ -47,6 +66,40 @@ class AppDialog : BottomSheetDialogFragment() {
...
+    private val documentPicker = registerForActivityResult(ActivityResultContracts.OpenDocument()) {
+        it?.let { uri ->
+            try {
+                val savesFolder = File(savesFolderRoot)
+                var inputZip = requireContext().contentResolver.openInputStream(uri)
+                var validZip = false
+                // A TitleID must be the first folder name inside the zip in order to be considered valid.
+                if (inputZip != null) {
+                    ZipInputStream(BufferedInputStream(inputZip)).use { zis ->
+                        validZip = Regex("^0100[0-9A-Fa-f]{12}/").containsMatchIn(zis.nextEntry.name)
```
AbandonedCart, line 80: 0-9 can be replaced with \d

app/src/main/java/emu/skyline/AppDialog.kt
```diff
@@ -101,6 +154,66 @@ class AppDialog : BottomSheetDialogFragment() {
...
+                }.show()
+        }
+
+        binding.importSave.setOnClickListener {
+            documentPicker.launch(arrayOf("application/zip"))
+        }
+
+        binding.exportSave.isEnabled = saveExists
+        binding.exportSave.setOnClickListener {
+            CoroutineScope(Dispatchers.IO).launch {
```
AbandonedCart, line 181: CoroutineScope(Dispatchers.IO).launch(Dispatchers.IO) resolves warnings about potential thread starvation

app/src/main/java/emu/skyline/AppDialog.kt
```diff
@@ -47,6 +66,40 @@ class AppDialog : BottomSheetDialogFragment() {
 
     private val item by lazy { requireArguments().serializable<AppItem>(AppItemTag)!! }
 
+    private val savesFolderRoot by lazy { "${requireContext().getPublicFilesDir().canonicalPath}/switch/nand/user/save/0000000000000000/00000000000000000000000000000001/" }
```
lynxnb, line 69: No need for lazy here, make this a val property.

app/src/main/java/emu/skyline/AppDialog.kt
lynxnb: Do this.

app/src/main/java/emu/skyline/AppDialog.kt
lynxnb: No need to do this, I don't even know why Android Studio would even warn about this here.

app/src/main/java/emu/skyline/AppDialog.kt
```diff
@@ -101,6 +154,66 @@ class AppDialog : BottomSheetDialogFragment() {
...
+                    dialogInterface.cancel()
+                }.show()
+        }
+
+        binding.importSave.setOnClickListener {
+            documentPicker.launch(arrayOf("application/zip"))
+        }
+
+        binding.exportSave.isEnabled = saveExists
+        binding.exportSave.setOnClickListener {
```
lynxnb, line 180: Definitely encapsulate this logic to a separate `exportSave` function, the `onViewCreated` method is already pretty long, no need to make it bigger.

app/src/main/java/emu/skyline/AppDialog.kt
```diff
@@ -101,6 +154,66 @@ class AppDialog : BottomSheetDialogFragment() {
...
+            CoroutineScope(Dispatchers.IO).launch {
+                try {
+                    val saveFolder = File(saveFolderPath)
+                    val outputZipFile = File("$savesFolderRoot${item.title} (v${binding.gameVersion.text}) [${item.titleId}] - ${LocalDateTime.now().format(DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm"))}.zip")
+                    lastZipCreated?.delete()
+                    outputZipFile.createNewFile()
+                    ZipOutputStream(BufferedOutputStream(FileOutputStream(outputZipFile))).use { zos ->
+                        saveFolder.walkTopDown().forEach { file ->
+                            val zipFileName = file.absolutePath.removePrefix(savesFolderRoot).removePrefix("/")
+                            if (zipFileName == "") return@forEach
```
lynxnb, line 190: Place the `return` on the following line.

app/src/main/java/emu/skyline/AppDialog.kt
```diff
@@ -101,6 +154,66 @@ class AppDialog : BottomSheetDialogFragment() {
...
+                            zos.putNextEntry(entry)
+                            if (file.isFile) {
+                                file.inputStream().use { fis -> fis.copyTo(zos) }
+                            }
+                        }
+                    }
+                    lastZipCreated = outputZipFile
+                } catch (e: Exception) {
+                    withContext(Dispatchers.Main){
+                        Snackbar.make(binding.root, e.message as CharSequence, Snackbar.LENGTH_LONG).show()
```
lynxnb, line 201: Showing Java exception messages is bad, and not useful to the user at all. Either enumerate all the possible failure points each with a proper error string, or show a generic failure message.

app/src/main/java/emu/skyline/AppDialog.kt
```diff
@@ -117,4 +230,9 @@ class AppDialog : BottomSheetDialogFragment() {
             }
         }
     }
+
+    override fun onDestroyView(){
+        super.onDestroyView()
+        lastZipCreated?.delete()
```
lynxnb, line 236: Don't keep a zip file laying around in the `savesFolderPath`, but use the app cache folder, in the same way the import function does. That way you don't need to destroy the file here, but you'd destroy it right after "returning" from the intent. See `startActivityForResult`.

app/src/main/res/values/strings.xml
```diff
@@ -263,4 +267,10 @@
     <string name="expand_button_title" tools:override="true">Expand</string>
     <string name="undo">Undo</string>
     <string name="per_game_settings_active_message">Per-game settings are active</string>
+    <string name="delete_save_confirmation_message">Are you sure you want to delete this save?</string>
+    <string name="action_irreversible">This action is irreversible</string>
+    <string name="yes">Yes</string>
+    <string name="no">No</string>
+    <string name="save_file_imported_ok">The save file was imported successfully</string>
+    <string name="save_file_invalid_zip_structure">Failed to unzip. First subfolder must be named after the TitleId of the game.</string>
```
lynxnb, line 275: `Invalid Zip directory structure: the first subfolder name must be the Title ID of the game.`

app/src/main/java/emu/skyline/AppDialog.kt
```diff
@@ -47,6 +66,40 @@ class AppDialog : BottomSheetDialogFragment() {
 
     private val item by lazy { requireArguments().serializable<AppItem>(AppItemTag)!! }
 
+    private val savesFolderRoot by lazy { "${requireContext().getPublicFilesDir().canonicalPath}/switch/nand/user/save/0000000000000000/00000000000000000000000000000001/" }
+    private var lastZipCreated : File? = null
+    private val documentPicker = registerForActivityResult(ActivityResultContracts.OpenDocument()) {
+        it?.let { uri ->
```
lynxnb, line 72: Just like the export function, I'd also make an `importSave` function here to avoid having all this logic in a variable that isn't strictly related to it. Also add documentation (KDoc) on both the export and import functions.

lynxnb (changes requested):
A few overall comments:
- Your use of coroutines is correct, good job on that
- The level of `if` nesting in your code makes it harder to read, also the lack of formatting/empty lines contributes to this. You can easily get rid of most of the nesting issues with early returns checking for error conditions.
- Remember to run the formatter on all code you PR, `Ctrl`+`Alt`+`L` IIRC is the default keybind on Android Studio

app/src/main/java/emu/skyline/AppDialog.kt
PabloG02: Forgot to remove it. It was done to debug an issue in discord 😅

app/src/main/java/emu/skyline/AppDialog.kt
AbandonedCart: It doesn't warn about it here. It warns that `createNewFile` and `FileOutputStream` are `possibly blocking calls in non-blocking context that could lead to thread starvation`. Android Studio actually wants you to wrap both in withContext to suspend the thread until the method is complete, so this is the "suppression" version.

app/src/main/java/emu/skyline/AppDialog.kt
```diff
@@ -117,4 +169,108 @@ class AppDialog : BottomSheetDialogFragment() {
...
+                }
+            }
+            lastZipCreated = outputZipFile
+        } catch (e : Exception) {
+            return false
+        }
+        return true
+    }
+
+    private val startForResultExportSave = registerForActivityResult(ActivityResultContracts.StartActivityForResult()) { _ ->
```
AbandonedCart, line 203: `_ ->` isn't needed here

app/src/main/java/emu/skyline/AppDialog.kt
```diff
@@ -117,4 +169,108 @@ class AppDialog : BottomSheetDialogFragment() {
...
+
+                withContext(Dispatchers.Main) {
+                    if (validZip) {
+                        val isSaveFileOfThisGame = File("$savesFolderRoot${item.titleId}").exists()
+                        binding.deleteSave.isEnabled = isSaveFileOfThisGame
+                        binding.exportSave.isEnabled = isSaveFileOfThisGame
+                        Snackbar.make(binding.root, R.string.save_file_imported_ok, Snackbar.LENGTH_LONG).show()
+                    } else {
+                        Snackbar.make(binding.root, getString(R.string.save_file_invalid_zip_structure), Snackbar.LENGTH_LONG).show()
+                    }
```
lynxnb, line 267:
```suggestion
                    if (!validZip) {
                        Snackbar.make(binding.root, getString(R.string.save_file_invalid_zip_structure), Snackbar.LENGTH_LONG).show()
                        return@withContext
                    }

                    val isSaveFileOfThisGame = File("$savesFolderRoot${item.titleId}").exists()
                    binding.deleteSave.isEnabled = isSaveFileOfThisGame
                    binding.exportSave.isEnabled = isSaveFileOfThisGame
                    Snackbar.make(binding.root, R.string.save_file_imported_ok, Snackbar.LENGTH_LONG).show()
```

app/src/main/java/emu/skyline/AppDialog.kt
```diff
@@ -117,4 +169,108 @@ class AppDialog : BottomSheetDialogFragment() {
...
+            ZipOutputStream(BufferedOutputStream(FileOutputStream(outputZipFile))).use { zos ->
+                saveFolder.walkTopDown().forEach { file ->
+                    val zipFileName = file.absolutePath.removePrefix(savesFolderRoot).removePrefix("/")
+                    if (zipFileName == "")
+                        return@forEach
+                    val entry = ZipEntry("$zipFileName${(if (file.isDirectory) "/" else "")}")
+                    zos.putNextEntry(entry)
+                    if (file.isFile) {
+                        file.inputStream().use { fis -> fis.copyTo(zos) }
+                    }
```
lynxnb, line 193: Single line `if`, drop the `{}`.

app/src/main/java/emu/skyline/AppDialog.kt
```diff
@@ -101,6 +126,33 @@ class AppDialog : BottomSheetDialogFragment() {
...
+        binding.deleteSave.isEnabled = saveExists
+        binding.deleteSave.setOnClickListener {
+            AlertDialog.Builder(requireContext())
+                .setTitle(getString(R.string.delete_save_confirmation_message))
+                .setMessage(getString(R.string.action_irreversible))
+                .setPositiveButton(getString(R.string.yes)) { dialogInterface, _ ->
+                    File(saveFolderPath).deleteRecursively()
+                    binding.deleteSave.isEnabled = false
+                    binding.exportSave.isEnabled = false
+                    dialogInterface.dismiss()
```
lynxnb, line 141: No need to dismiss manually, the framework already does so after handling positive/negative buttons.

app/src/main/java/emu/skyline/AppDialog.kt
```diff
@@ -101,6 +126,33 @@ class AppDialog : BottomSheetDialogFragment() {
...
+                }.setNegativeButton(getString(R.string.no)) { dialogInterface, _ ->
+                    dialogInterface.cancel()
+                }.show()
```
lynxnb, line 144: Same as above, no need to cancel manually:
```suggestion
                }
                .setNegativeButton(getString(R.string.no), null)
                .show()
```

app/src/main/java/emu/skyline/AppDialog.kt
```diff
@@ -117,4 +169,108 @@ class AppDialog : BottomSheetDialogFragment() {
...
+        CoroutineScope(Dispatchers.IO).launch {
+            if (!zipSave(saveFolderPath)) {
+                withContext(Dispatchers.Main) {
+                    Snackbar.make(binding.root, R.string.error, Snackbar.LENGTH_LONG).show()
+                }
+                return@launch
+            }
+
+            withContext(Dispatchers.Main) {
+                val file = DocumentFile.fromSingleUri(requireContext(), DocumentsContract.buildDocumentUri(DocumentsProvider.AUTHORITY, "${DocumentsProvider.ROOT_ID}/switch/nand/user/save/0000000000000000/00000000000000000000000000000001/${lastZipCreated!!.name}"))!!
```
lynxnb, line 221: `lastZipCreated!!` doesn't look too good. Introduce a local variable of type `File` (non-nullable), where you store a temporary copy of `lastZipCreated`. You assign it above if `zipSave` returns `true`.

See the above suggestion, use `lastZipFile` in here instead of `lastZipCreated`.

app/src/main/java/emu/skyline/AppDialog.kt
```diff
@@ -117,4 +169,108 @@ class AppDialog : BottomSheetDialogFragment() {
...
+     * @param saveFolderPath The path to the folder containing the save file(s) to export.
+     */
+    private fun exportSave(saveFolderPath : String) {
+        CoroutineScope(Dispatchers.IO).launch {
+            if (!zipSave(saveFolderPath)) {
+                withContext(Dispatchers.Main) {
+                    Snackbar.make(binding.root, R.string.error, Snackbar.LENGTH_LONG).show()
+                }
+                return@launch
+            }
```
lynxnb, line 218:
```suggestion
            val lastZipFile = lastZipCreated

            if (!zipSave(saveFolderPath) || lastZipFile == null) {
                withContext(Dispatchers.Main) {
                    Snackbar.make(binding.root, R.string.error, Snackbar.LENGTH_LONG).show()
                }
                return@launch
            }
```

lynxnb (changes requested):

lynxnb (approved):

PR #2295, opened 2023-04-07
Remove header rom filter

dima-xd:
People were saying it's useless. Not sure if I did it correctly

PR #2312, opened 2023-04-17
Remove header rom filter (append)

AbandonedCart:
This is the extension of #2295 discussed in that PR


AbandonedCart:
Does the sorting itself stay as an option or should that be stripped?

lynxnb:
IMO, we get rid of it. Has it been any useful to anyone ever?

lynxnb:
Also I quickly glanced at the code, and I can say for sure that much more can be done in removing the filtering code. We load roms in a per-format map in `RomProvider.kt`, that should go as well.

AbandonedCart:
Removing anything else was waiting on the previous decision because a lot of it was tied to the sorting. There would have been no point trying to retain the sorting and then end up removing it anyway.

AbandonedCart:
That last one could be replaced with actually setting the serialVersionUID, but I wasn't sure how you wanted to come up with the value.

lynxnb:
SerialVersionUID doesn't matter for `AppItem` so it's fine just silencing the warning. I'll be reviewing the rest soon. Have you tried running the app and seeing if it works? Remember to try with a clean install of skyline too.

lynxnb (approved):
LGTM, great job, also to @dima-xd!
As a side note, for future reference, you don't need to remove strings from all languages, that is done automatically by Transifex when we sync translations. You only need to remove them from the main `strings.xml`.

AbandonedCart:
I just realized I never answered. Yes, I tested it. The one issue I found is in the PR for obscure crashes

Sugiyaro:
Huss

PR #2320, opened 2023-04-19
Crash fixes for uncommon occurences 

AbandonedCart:
Pause can happen in a few ways and some may not retain a flag that the emulator was paused, which can result in resume not being fired. What little is saved by not changing settings that are already set would be lost trying to effectively track the specific method the emulator was paused.

### DIFF
--- a/app/src/main/java/emu/skyline/utils/ParcelableHelper.kt
+++ b/app/src/main/java/emu/skyline/utils/ParcelableHelper.kt
@@ -15,13 +15,11 @@ inline fun <reified T : Parcelable> Bundle.parcelable(key: String): T? = when {
     else -> @Suppress("DEPRECATION") getParcelable(key) as? T
 }
 
-@Suppress("unused")
 inline fun <reified T : Parcelable> Intent.parcelable(key: String): T? = when {
     Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU -> getParcelableExtra(key, T::class.java)
     else -> @Suppress("DEPRECATION") getParcelableExtra(key) as? T
 }
 
-@Suppress("unused")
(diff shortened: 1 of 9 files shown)


app/src/main/java/emu/skyline/utils/ParcelableHelper.kt
```diff
@@ -0,0 +1,28 @@
+/*
+ * SPDX-License-Identifier: MPL-2.0
+ * Copyright © 2023 Skyline Team and Contributors (https://github.com/skyline-emu/)
+ */
+
+@file:Suppress("unused")
```
lynxnb, line 6: You can remove this suppress, we don't hate unused warnings.

app/src/main/java/emu/skyline/utils/ParcelableHelper.kt
```diff
@@ -0,0 +1,28 @@
...
+
+inline fun <reified T : Parcelable> Intent.parcelable(key: String): T? = when {
+    Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU -> getParcelableExtra(key, T::class.java)
+    else -> @Suppress("DEPRECATION") getParcelableExtra(key) as? T
+}
+
+inline fun <reified T : Parcelable> Intent.parcelableArrayList(key: String): ArrayList<T>? = when {
+    Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU -> getParcelableArrayListExtra(key, T::class.java)
+    else -> @Suppress("DEPRECATION") getParcelableArrayListExtra(key)
+}
```
lynxnb, line 28: New line at the end of the file.

lynxnb (commented):
Looks good, but could you drop the `Prevent a delayed coroutine from crashing` commit? I'm already fixing that myself.

lynxnb (approved):

PR #2326, opened 2023-04-20
Prevent a delayed coroutine from crashing

AbandonedCart:
> Checking state when the function is called only works if the coroutine is fired immediately. If it is delayed by another process, it will crash with a ClassCastException


AbandonedCart:
Closing this in favor of rolling a few crash and resume fixes together.

Issue #1923, opened 2022-10-25
Dual Screen mode with virtual controller.


mstrblueskys:
If I'm playing on a dual screen device, I don't have the option to play in device portrait, or 'dual landscape' mode. I'd love to be able to set the top screen as my play screen and the bottom as my virtual controller

PR #2328, opened 2023-04-21
Added folding devices support

skriand:
Fix to issue #1923
When app is spanned on dual-screen device (Surface Duo) or folding device is folded 90 degrees, game view is located in center of upper separate section.

### DIFF
--- a/app/src/main/java/emu/skyline/EmulationActivity.kt
+++ b/app/src/main/java/emu/skyline/EmulationActivity.kt
@@ -504,27 +504,26 @@ class EmulationActivity : AppCompatActivity(), SurfaceHolder.Callback, View.OnTo
     * Updating the layout depending on type and state of device
     */
     private fun updateCurrentLayout(newLayoutInfo: WindowLayoutInfo) {
-        val displayMetrics = DisplayMetrics()
-        windowManager.defaultDisplay.getMetrics(displayMetrics)
-        binding.onScreenGameView.minimumHeight = displayMetrics.heightPixels
-        requestedOrientation =
-            if (emulationSettings.orientation != ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED) {
-                emulationSettings.orientation
+        if (!emulationSettings.supportFoldableScreen) return
+        binding.onScreenGameView.minimumHeight = with (windowManager) {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+                maximumWindowMetrics.bounds.height()
             } else {
-                ActivityInfo.SCREEN_ORIENTATION_SENSOR_LANDSCAPE
+                val displayMetrics = DisplayMetrics()
+                @Suppress("DEPRECATION")
+                defaultDisplay.getMetrics(displayMetrics)
+                displayMetrics.heightPixels
             }
-        for (displayFeature in newLayoutInfo.displayFeatures) {
-            val foldFeature = displayFeature as? FoldingFeature
-            foldFeature?.let {
-                //Folding feature separates the display area into two distinct sections
-                if (it.isSeparating) {
-                    if (emulationSettings.orientation == ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED){
-                        requestedOrientation = emulationSettings.orientation
-                    }
-                    if (it.orientation == FoldingFeature.Orientation.HORIZONTAL) {
-                        binding.onScreenGameView.minimumHeight = displayFeature.bounds.top
-                    }
-                }
+        }
+        requestedOrientation = emulationSettings.orientation
+        val foldingFeature = newLayoutInfo.displayFeatures.find { it is FoldingFeature }
+        (foldingFeature as? FoldingFeature)?.let {
+            if (it.isSeparating) {
+                requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED
+                if (it.orientation == FoldingFeature.Orientation.HORIZONTAL)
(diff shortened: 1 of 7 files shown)


AbandonedCart:
It looks like this is only supporting one specific type of hinge, but disabling the user orientation on both. To put it another way, it would work on the Flip, but on the Fold it would only ignore the user orientation setting.

skriand:
[Documentation](https://learn.microsoft.com/en-us/dual-screen/android/jetpack/window-manager/) says, that property `isSeparating` for dual-screen devices will always be true when an app is spanned across the hinge. For other foldable devices, this will only be true when the state is `HALF_OPENED`, like when a device is in the tabletop position.
So, it shouldn't disturb on the Fold.

AbandonedCart:
I think you misunderstood.

>     if (it.isSeparating) { <- this determines the device is bent
>         //Folding feature separates the display area into two distinct sections
>         requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED <- this ignores user orientation
>         if (foldFeature.orientation == FoldingFeature.Orientation.HORIZONTAL) { <- this checks the hinge axis
>                 binding.onScreenGameView.minimumHeight = displayFeature.bounds.top
>         }
>     }

The Fold is talking about the [Galaxy Z Fold](https://www.samsung.com/us/smartphones/galaxy-z-fold4/).

Per the [Android documentation](https://developer.android.com/guide/topics/large-screens/make-apps-fold-aware), [FoldingFeature.Orientation.HORIZONTAL](https://developer.android.com/reference/androidx/window/layout/FoldingFeature.Orientation#HORIZONTAL()) is one of two [FoldingFeature.Orientation](https://developer.android.com/reference/androidx/window/layout/FoldingFeature.Orientation) types. 

The code, as it is, sets the orientation to SCREEN_ORIENTATION_UNSPECIFIED for any FoldingFeature.Orientation, but only sets the bounds when the FoldingFeature.Orientation is FoldingFeature.Orientation.HORIZONTAL 

Hopefully this makes more sense.

AbandonedCart:
Wouldn't it be easier to move the original orientation override inside the condition and then add a simple true / false option to enable folding to reposition the screen?

>    private fun updateCurrentLayout(newLayoutInfo: WindowLayoutInfo) {
>        if (!appSettings.enableFolding) return

...

>     if (it.isSeparating) {
>         if (foldFeature.orientation == FoldingFeature.Orientation.HORIZONTAL) {
>                 requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED
>                 binding.onScreenGameView.minimumHeight = displayFeature.bounds.top
>         }
>     }

skriand:
Yes, it's really easier that way.

AbandonedCart:
You'd probably also want to get the height `by lazy` outside the function to spare a little memory. I didn't think of that until after, but any of that you want to use is yours.

AbandonedCart:
```
git reset --hard 4c7fd0447d7333d815d2d9e8343e5f97ae9a24d1
git cherry-pick cb6f5c5d47e274631831abdf56372df71a15db34 06635d7f69a2e2d09e60fac60d1799aa466a024c 0b71570661ac080fdc92412f0431bf6a4001b60f 79e49e03a0addd3e3cd7660e727a12c8ff239b00
git push -f
```

This will remove all of the merge messages and leave just the commits. The hashes are based on your repo.

AbandonedCart:
I made some adjustments to move the pause button and performance stats into the bottom screen. Along with that, I added a bit of a buffer to the top and bottom of the bottom screen to avoid placing buttons in the bend and having the controls at the edge of the screen. 

I can also submit this separately later to avoid it being something you have to get approved as part of this PR, since it does not replace any of the code already included. I figured I would mention it in case you wanted it, though.

https://github.com/8bitDream/skyline/commit/68a3f95abed351f5dcb7178c1efb01ebb0f1ff80

bylaws:
@lynx

bylaws:
@lynxnb

lynxnb:
I'll get to this soon.

lynxnb:
Also, I'm not sure `Support Foldable Screens` is that good of a name, but can't think of a better one, so I guess we'll keep that. Suggestions are welcome.

AbandonedCart:
Samsung calls it Flex Mode, but that doesn't cover OPPO or Microsoft. 

Split screen sounds like the feature on any screen tall / wide enough.

The only real alternative would be using "display" instead of "screen"

app/src/main/java/emu/skyline/EmulationActivity.kt
```diff
@@ -480,6 +499,24 @@ class EmulationActivity : AppCompatActivity(), SurfaceHolder.Callback, View.OnTo
...
+        binding.onScreenGameView.layoutParams.height = ViewGroup.LayoutParams.MATCH_PARENT
+        requestedOrientation = emulationSettings.orientation
+        val foldingFeature = newLayoutInfo.displayFeatures.find { it is FoldingFeature }
+        (foldingFeature as? FoldingFeature)?.let {
+            if (it.isSeparating) {
+                requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED
+                if (it.orientation == FoldingFeature.Orientation.HORIZONTAL)
+                    binding.onScreenGameView.layoutParams.height = it.bounds.top
+            }
+        }
```
lynxnb, line 516: Would love to see some comments in here to explaining the **why** of what is being done. Unlikely anyone will remember months from now, let alone years in the future.

nickbeth (changes requested):

PR #2330, opened 2023-04-24
Graphics Pipeline Loading UI

nickbeth:
Add functionality for the UI screen from #2255 

Credits to @PabloG02 for the XML layout.

### DIFF
--- a/app/src/main/cpp/skyline/gpu/graphics_pipeline_assembler.cpp
+++ b/app/src/main/cpp/skyline/gpu/graphics_pipeline_assembler.cpp
@@ -214,6 +214,9 @@ namespace skyline::gpu {
 
         std::scoped_lock lock{mutex};
         compilePendingDescs.erase(pipelineDescIt);
+        if (compilationCallback)
+            compilationCallback();
+
         return pipeline;
     }
 
@@ -251,5 +254,16 @@ namespace skyline::gpu {
             std::vector<u8> rawData{vkPipelineCache.getData()};
             SerialisePipelineCache(gpu, pipelineCacheDir, rawData);
         });
+    }
+
+    void GraphicsPipelineAssembler::RegisterCompilationCallback(std::function<void()> callback) {
+        if (compilationCallback)
+            throw exception("A compilation callback is already registered");
+
+        compilationCallback = std::move(callback);
+    }
+
+    void GraphicsPipelineAssembler::UnregisterCompilationCallback() {
+        compilationCallback = {};
     }
 }
--- a/app/src/main/cpp/skyline/gpu/interconnect/maxwell_3d/pipeline_manager.cpp
+++ b/app/src/main/cpp/skyline/gpu/interconnect/maxwell_3d/pipeline_manager.cpp
@@ -9,6 +9,7 @@
 #include <gpu/graphics_pipeline_assembler.h>
 #include <gpu/shader_manager.h>
 #include <gpu.h>
+#include <jvm.h>
 #include <vulkan/vulkan_enums.hpp>
 #include "graphics_pipeline_state_accessor.h"
 #include "pipeline_manager.h"
@@ -942,12 +943,19 @@ namespace skyline::gpu::interconnect::maxwell3d {
         });
     }
 
-    PipelineManager::PipelineManager(GPU &gpu) {
+    PipelineManager::PipelineManager(GPU &gpu, JvmManager &jvm) {
         if (!gpu.graphicsPipelineCacheManager)
             return;
 
-        std::ifstream stream{gpu.graphicsPipelineCacheManager->OpenReadStream()};
+        std::atomic<u32> compiledCount{};
+        auto [stream, totalPipelineCount]{gpu.graphicsPipelineCacheManager->OpenReadStream()};
         i64 lastKnownGoodOffset{stream.tellg()};
+
+        jvm.ShowPipelineLoadingScreen(totalPipelineCount);
+        gpu.graphicsPipelineAssembler->RegisterCompilationCallback([&]() {
+            jvm.UpdatePipelineLoadingProgress(++compiledCount);
+        });
+
         try {
             auto startTime{util::GetTimeNs()};
             PipelineStateBundle bundle;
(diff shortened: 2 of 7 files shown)


app/src/main/cpp/skyline/gpu/graphics_pipeline_assembler.cpp
```diff
@@ -252,4 +255,8 @@ namespace skyline::gpu {
             SerialisePipelineCache(gpu, pipelineCacheDir, rawData);
         });
     }
+
+    void GraphicsPipelineAssembler::RegisterCompilationCallback(std::function<void()> callback) {
+        compilationCallback = std::move(callback);
+    }
```
bylaws, line 261: Make unregistering a seperate function, and throw exception here if a callback is already registered

app/src/main/cpp/skyline/gpu/interconnect/maxwell_3d/pipeline_manager.cpp
```diff
@@ -942,12 +943,20 @@ namespace skyline::gpu::interconnect::maxwell3d {
...
     }
 
-    PipelineManager::PipelineManager(GPU &gpu) {
+    PipelineManager::PipelineManager(GPU &gpu, JvmManager &jvm) {
         if (!gpu.graphicsPipelineCacheManager)
             return;
 
-        std::ifstream stream{gpu.graphicsPipelineCacheManager->OpenReadStream()};
+        u32 pipelineCount{};
+        jint compiledCount{};
```
bylaws, line 951: Don't use java types in GPU code, this also needs to be atomic. 

std::atomic<i32> currentCompiledPipelineCount{};

app/src/main/cpp/skyline/gpu/interconnect/maxwell_3d/pipeline_manager.cpp
```diff
@@ -942,12 +943,20 @@ namespace skyline::gpu::interconnect::maxwell3d {
...
+        std::ifstream stream{gpu.graphicsPipelineCacheManager->OpenReadStream(&pipelineCount)};
```
bylaws, line 952: Have this return an std pair, and then use structured binding as approved. Also name like totalPipelineCount or smth

app/src/main/cpp/skyline/gpu/pipeline_cache_manager.cpp
```diff
@@ -3,24 +3,32 @@
 
 #include <ostream>
 #include <os.h>
+#include "interconnect/kepler_compute/pipeline_manager.h"
```
bylaws, line 6: Idk why you're doing this but don't

app/src/main/cpp/skyline/gpu/pipeline_cache_manager.cpp
```diff
@@ -30,8 +38,16 @@ namespace skyline::gpu {
             writeCondition.wait(lock, [this] { return !writeQueue.empty(); });
             auto bundle{std::move(writeQueue.front())};
             writeQueue.pop();
+            header.count++;
```
bylaws, line 41: You can move this to below, just before the write

bylaws (changes requested):

app/src/main/cpp/skyline/gpu/pipeline_cache_manager.cpp
lynxnb: Android Studio sneaked that in

app/src/main/cpp/skyline/gpu/pipeline_cache_manager.cpp
```diff
@@ -3,24 +3,32 @@
...
 
 namespace skyline::gpu {
     struct PipelineCacheFileHeader {
         static constexpr u32 Magic{util::MakeMagic<u32>("PCHE")}; //!< The magic value used to identify a pipeline cache file
-        static constexpr u32 Version{2}; //!< The version of the pipeline cache file format, MUST be incremented for any format changes
+        static constexpr u32 Version{3}; //!< The version of the pipeline cache file format, MUST be incremented for any format changes
 
         u32 magic{Magic};
         u32 version{Version};
+        u32 count{0}; //!< The total number of pipeline cache bundles in the file
```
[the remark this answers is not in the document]
lynxnb: I like the explicitness of {0}, can I keep it that way?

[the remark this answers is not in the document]
bylaws: Whatever sure

app/src/main/cpp/skyline/gpu/graphics_pipeline_assembler.cpp
lynxnb: IMO we should add an override boolean parameter defaulted to false instead of throwing an exception

app/src/main/cpp/skyline/gpu/interconnect/maxwell_3d/pipeline_manager.cpp
lynxnb: Does it really need to be atomic? It's an 32bit int, it's going to be atomic in any case, and even if it isn't we don't really care, do we?

lynxnb: I made it atomic

app/src/main/cpp/skyline/gpu/graphics_pipeline_assembler.cpp
bylaws: Nah, that's confusing

app/src/main/cpp/skyline/gpu/interconnect/maxwell_3d/pipeline_manager.cpp
bylaws: > Does it really need to be atomic? It's an 32bit int, it's going to be atomic in any case, and even if it isn't we don't really care, do we?

We aren't x86, memory accesses are not generally atomic on arm

app/src/main/cpp/skyline/gpu/graphics_pipeline_assembler.cpp
lynxnb: Ok

PR #2336, opened 2023-05-02
Game folders (DLC / Updates / Etc)

AbandonedCart:
This is mostly groundwork for future DLC / Update support. Right now, though, it allows those folders to be hidden in the browser without having to keep them isolated outside of the folder used by the app. 

In the future, it can be repurposed as the base folder for retrieving those files for the specific game, similar to the DLC management interface in some of the desktop emulators.


bylaws:
When we add dlc support, we'll probably have the game folder store them as well. As such I am not sure this really has any point, and it may confuse users adding it now as DLC and updates aren't supported. Lmk if you can see any other benefits

AbandonedCart:
Closing in favor of #2345, but I'll leave the branch in case DLC support ends up similar to desktop implementations. The ability to store a map with array values as as a preference may end up being useful. Even if the files are stored alongside games, it will likely require the ability to optionally enable or disable individual DLC.

PR #2345, opened 2023-05-05
Add an option to filter files parsed as invalid

AbandonedCart:
This allows the game list to (optionally) skip displaying files that were parsed as invalid. It can filter out junk files kept in the game folder and DLC, which also parse as invalid. Again, it is optional. At such time as DLC are no longer parsed as invalid, this would automatically stop applying to them and it will be up to the emulator to determine how to handle that.

This is the third and final attempt to balance the convenience of storing all files together without a negative impact on navigating the emulator efficiently. This is a compromise between not making any assumptions about the file that the emulator hasn't already made and not setting any expectations on how DLC will be supported.

<details>
  <summary>Before / Disabled</summary>

signal-2023-05-06-072601_002
</details>

<details>
  <summary>After / Enabled</summary>

signal-2023-05-06-072604_002
</details>

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -59,6 +59,8 @@
     <string name="select_action">Always Show Game Information</string>
     <string name="select_action_desc_on">Game information will be shown on clicking a game</string>
     <string name="select_action_desc_off">Game information will only be shown on long-clicking a game</string>
+    <string name="filter_invalid_files_desc_off">Show files that have been parsed as invalid</string>
+    <string name="filter_invalid_files_desc_on">Hide files that have been parsed as invalid</string>
     <!-- Settings - Game -->
     <string name="game">Game</string>
     <string name="use_custom_settings">Enable Custom Settings</string>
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -59,6 +59,9 @@
     <string name="select_action">Always Show Game Information</string>
     <string name="select_action_desc_on">Game information will be shown on clicking a game</string>
     <string name="select_action_desc_off">Game information will only be shown on long-clicking a game</string>
+    <string name="filter_invalid_files">Hide Invalid Files</string>
+    <string name="filter_invalid_files_desc_off">Show files that have been parsed as invalid</string>
```
nickbeth, line 62: Would probably drop this string and only keep one for the description.

nickbeth (approved):
FWIW, this is a great idea, really like it.

app/src/main/res/values/strings.xml
AbandonedCart: I didn't like the title, so that works.

PR #2346, opened 2023-05-05
Export custom settings to text

ismaelestalayo:
Created a very simple function to export a game's custom settings (or global settings if custom are disabled) to text, so users can then easily share their configs in Discord (the button placement can be seen in the background of the screenshot as well):

image

If you are not a fan of ascii emojis, i could always just replace then by ON/OFF as well

### DIFF
--- a/app/src/main/java/emu/skyline/preference/ExportCustomSettingsPreference.kt
+++ b/app/src/main/java/emu/skyline/preference/ExportCustomSettingsPreference.kt
@@ -1,0 +1,78 @@
+/*
+ * SPDX-License-Identifier: MPL-2.0
+ * Copyright © 2023 Skyline Team and Contributors (https://github.com/skyline-emu/)
+ */
+
+package emu.skyline.preference
+
+import android.content.Context
+import android.util.AttributeSet
+import androidx.preference.Preference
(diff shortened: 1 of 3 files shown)


app/src/main/java/emu/skyline/preference/ExportCustomSettingsPreference.kt
```diff
@@ -48,15 +48,16 @@ class ExportCustomSettingsPreference @JvmOverloads constructor(context : Context
                 - Docked: $systemIsDocked
                 
                 GPU
-                - Driver: $gpuDriver (executors: $gpuExecSlotCount slots, threshold of $gpuExecFlushThreshold)
+                - Driver: $gpuDriver
+                - Executors: $gpuExecSlotCount slots, threshold of $gpuExecFlushThreshold)
```
AbandonedCart, line 52: There is no `(` to go along with the `)`

ismaelestalayo: fixed!

nickbeth (commented):
FWIW now, I would have left some comments on your code, but overall your contribution was fine.

